PR #10, opened 2022-09-11
ic2 support (exp/classic)

ISurrealI:


brachy84:
Ok now you are logging error for recipe removal, but since you didnt add builders, you also need to log errors on addition

brachy84:
Also please add recipe streams

brachy84 (approved):

PR #17, opened 2022-11-02
Rewrite script remapping

brachy84:
Before i generated the file myself, which was not very good.
Now i use a official mappings file where i deleted all the useless class mappings.
This also fixes remapping for subclasses. This is done at runtime by getting the super class if the field/method couldn't be found in the current class.
Im not sure if caching everything (for every class and sub class) would be better


Rongmario (approved):
Looks good to me. Ready to merge.

PR #50, opened 2023-04-12
Extended Crafting compat

WaitingIdly:
what this PR adds:
- compat with Extended Crafting Compression crafting (crafttweaker level, consume catalyst)
- compat with Extended Crafting Combination crafting (crafttweaker level)
- compat with Extended Crafting Ender Crafting crafting (crafttweaker level, shapeless crafts, mirrored shaped crafts, key matrix)
- compat with Extended Crafting Table Crafting crafting (basic crafting, tier selection, shapeless/shaped/mirrored, key matrix)
- any of the above compats which have a config value for a default value respects that config
- a NonNullList conversion to IngredientHelper, and updated blood magic compat to use it
- an override in `IIngredient` for `withStack`

currently not implemented:
- recipe transformers for Ender Crafting (might not work regardless) and Table Crafting (part of the crafttweaker
- ability to use `null` in ingredient matrices, eg 
```groovy
mods.extendedcrafting.endercrafting.shapedBuilder()
        .output(item('minecraft:diamond') * 32)
        .matrix([[item('minecraft:gold_ingot'),null,item('minecraft:gold_ingot')],
                [null,item('minecraft:gold_ingot'),null],
                [item('minecraft:gold_ingot'),null,item('minecraft:gold_ingot')]])
        .time(1)
        .register()
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -116,6 +116,14 @@ dependencies {
         deobfProvided 'curse.maven:blood-magic-224791:2822288'
     }
 
+    if (project.debug_extended_crafting.toBoolean()) {
+        deobfCompile 'curse.maven:cucumber-272335:2645867'
+        deobfCompile 'curse.maven:extended-crafting-nomifactory-edition-398267:3613140'
+    } else {
+        deobfProvided 'curse.maven:cucumber-272335:2645867'
+        deobfProvided 'curse.maven:extended-crafting-nomifactory-edition-398267:3613140'
+    }
+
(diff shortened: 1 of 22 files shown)


brachy84:
I impoved recipe matching by implementing our own recipe class which just wraps a crafting recipe since the matching logic is exactly the same. Now stuff like transformers and match conditions should also work.
I also implemented that the recipe maker item will now generate groovy code instead of zs via mixin.

brachy84 (approved):

PR #56, opened 2023-05-20
Botania Example Script

Testure:
Example script for the Botania API.

Also fixes a warning in `ElvenTrade`.


src/main/java/com/cleanroommc/groovyscript/compat/mods/botania/ElvenTrade.java
```diff
@@ -62,7 +62,7 @@ public boolean removeByOutputs(ItemStack... outputs) {
         })) return true;
 
         GroovyLog.msg("Error removing Botania Elven Trade recipe")
-                .add("could not find recipe with outputs {}", outputs)
+                .add("could not find recipe with outputs {}", (Object[]) outputs)
```
brachy84, line 65: Arrays are not properly converted to string. Use `Arrays.toString(outputs)`.

brachy84 (changes requested):

examples/botania.groovy
WaitingIdly, line 1: belatedly: all of our docs include single quotes (`'`) for strings. this should be changed to also be the case here

WaitingIdly (changes requested):

WaitingIdly (approved):

examples/botania.groovy
brachy84: You can make a PR to change all of them if you want. I did it because i think it looks nicer.

brachy84 (approved):

PR #60, opened 2023-06-18
Add String based alternatives to builders that use Aspects

Testure:
Allows you to use `.aspect("aer", 5)` instead of `.aspect(aspect("aer") * 5)` for convenience.


WaitingIdly (changes requested):
could you add these new options to the example file?

WaitingIdly (approved):

brachy84 (changes requested):
Unlike the aspect bracket handler, this doesnt do any validation if the aspect really exists.

brachy84:
Now you completely changed the aspect bracket handler. An exception should only be thrown if there's an invalid amount of arguments or if the arguments have the wrong type. Otherwise it should return null. The bracket handler will then automatically handle the error message.

brachy84 (approved):

PR #63, opened 2023-06-23
Run PostInit scripts on LoadComplete

Testure:
Runs the `POST_INIT` loader on `FMLLoadComplete` instead of `FMLPostInitialization`. (while making sure run before JEI loads)

This solves any issues with scripts being run before a mod loads required recipes or other things, including an existing issue where Tinkers' registers melting recipes after scripts are run.

Also added logs that say how long the first script loads took.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/core/mixin/LoaderControllerMixin.java
+++ b/src/main/java/com/cleanroommc/groovyscript/core/mixin/LoaderControllerMixin.java
@@ -16,11 +16,7 @@ public class LoaderControllerMixin {
         if (state == LoaderState.PREINITIALIZATION) {
             GroovyScript.initializeGroovyPreInit();
         }
-    }
-
-    @Inject(method = "distributeStateMessage(Lnet/minecraftforge/fml/common/LoaderState;[Ljava/lang/Object;)V", at = @At("RETURN"))
-    public void postInit(LoaderState state, Object[] eventData, CallbackInfo ci) {
-        if (state == LoaderState.POSTINITIALIZATION) {
+        if (state == LoaderState.AVAILABLE) {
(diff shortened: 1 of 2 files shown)


brachy84:
I would like this to be consistent with pre init loading. See [LoaderControllerMixin](https://github.com/CleanroomMC/GroovyScript/blob/master/src/main/java/com/cleanroommc/groovyscript/core/mixin/LoaderControllerMixin.java).

src/main/java/com/cleanroommc/groovyscript/core/mixin/LoaderControllerMixin.java
```diff
@@ -17,4 +17,11 @@ public void preInit(LoaderState state, Object[] eventData, CallbackInfo ci) {
             GroovyScript.initializeGroovyPreInit();
         }
     }
+
+    @Inject(method = "distributeStateMessage(Lnet/minecraftforge/fml/common/LoaderState;[Ljava/lang/Object;)V", at = @At("RETURN"))
+    public void postInit(LoaderState state, Object[] eventData, CallbackInfo ci) {
+        if (state == LoaderState.POSTINITIALIZATION) {
+            GroovyScript.initializeGroovyPostInit();
+        }
+    }
```
brachy84, line 26: Why do create a new method and not use the exisiting mixin?

Testure: Because the existing method is at HEAD, which would cause scripts to run before fml post init

brachy84: Then use `LoaderState.AVAILABLE`

brachy84 (approved):

PR #68, opened 2023-06-25
Avaritia compat

brachy84:


src/main/java/com/cleanroommc/groovyscript/compat/mods/avaritia/Avaritia.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.cleanroommc.groovyscript.compat.mods.ModPropertyContainer;
+
+public class Avaritia extends ModPropertyContainer {
+
+    public final ExtremeCrafting extremeCrafting = new ExtremeCrafting();
+    public final Compressor compressor = new Compressor();
+
+    public Avaritia() {
+        addRegistry(this.extremeCrafting);
```
WaitingIdly, line 11: we dont do `this.` in other ModPropertyContainers

src/main/java/com/cleanroommc/groovyscript/compat/mods/avaritia/Compressor.java
WaitingIdly, line 1: where is the builder, removeAll, and streamRecipes?

build.gradle
```diff
@@ -108,6 +108,12 @@ dependencies {
 
     implementation 'mezz.jei:jei_1.12.2:4.16.1.302'
 
+    compileOnly rfg.deobf('curse.maven:avaritia_1_10-261348:3143349')
+    if (project.debug_avaritia.toBoolean()) {
+        runtimeOnly rfg.deobf('curse.maven:codechicken_lib_1_8-242818:2779848')
```
WaitingIdly, line 113: "probably not advised" - on having the same mod be `runtimeOnly`d twice

examples/avaritia.groovy
```diff
@@ -0,0 +1,15 @@
+
+mods.avaritia.ExtremeCrafting.removeByOutput(item('avaritia:resource', 6))
+mods.avaritia.ExtremeCrafting.shapedBuilder()
```
WaitingIdly, line 3: should get examples for a shapeless builder, string array and key replacement, and recipe functions/actions

gradle.properties
```diff
@@ -17,19 +17,20 @@ coremod_plugin_class_name = com.cleanroommc.groovyscript.core.GroovyScriptCore
 
 # Debug mod compat
 debug_actually_additions = false
+debug_astral = false
+debug_avaritia = true
```
WaitingIdly, line 21: false

src/main/java/com/cleanroommc/groovyscript/compat/mods/avaritia/ExtremeRecipeBuilder.java
WaitingIdly, line 1: the registry location is never set, defaulting to `null`, so only the last extreme crafting recipe appears in-game.

src/main/java/com/cleanroommc/groovyscript/compat/mods/avaritia/Compressor.java
```diff
@@ -0,0 +1,115 @@
...
+                .input(input)
+                .output(output)
+                .register();
+    }
+
+    public class RecipeBuilder extends AbstractRecipeBuilder<ICompressorRecipe> {
+
+        private int cost = 300;
+
+        public RecipeBuilder cost(int cost) {
```
WaitingIdly, line 81: for `ExtendedCrafting.CompressionCrafting.RecipeBuilder` i added another way to get the item amount - consider adding the same here?

WaitingIdly (changes requested):

src/main/java/com/cleanroommc/groovyscript/compat/mods/avaritia/Compressor.java
WaitingIdly: this might have been a bad idea, actually - since now, if `inputCount(100).input(item)` the inputCount will be overridden to be 1, making the output of the builder dependent on the order elements are built - definitely undesired.

brachy84: How about only setting input count if ingredient amount is > 1

WaitingIdly: a reasonable solution. would you mind also applying this to `CompressionCrafting.RecipeBuilder#input`?

WaitingIdly (approved):

PR #69, opened 2023-06-26
GHA Script

KorewaLidesu:
This PR mainly for new GHA that allow test artifact to be built when new commit created, include cache for faster build time.
Currently ignore release GHA file, LICENSE and README.md to avoid unneccessary build task.

P/s: Please squash this PR instead of merging it directly

### DIFF
--- a/.github/workflows/TestBuild.yaml
+++ b/.github/workflows/TestBuild.yaml
@@ -29,12 +29,13 @@ jobs:
         file: gradle.properties
         all: true
 
-    - name: Change mod version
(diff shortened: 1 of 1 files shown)


brachy84:
What do you mean with smash? :thonk: Do you mean squash?
Also, can i just merge it or does it something else?

KorewaLidesu:
> What do you mean with smash? :thonk: Do you mean squash? Also, can i just merge it or does it something else?

You can merge it but the commit just not worth, better squash it instead.
And yes i'm using wrong word 💀

PR #70, opened 2023-06-29
Add a varargs variant for yeeting items in JEI

Testure:
Allows you to yeet a ton of items at the same time via varargs, which is both more convenient and more performant.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/JustEnoughItems.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/JustEnoughItems.java
@@ -53,20 +53,14 @@ public class JustEnoughItems extends ModPropertyContainer {
             }
             JeiPlugin.hideItem(ingredient.getMatchingStacks());
         }
-        List<ResourceLocation> recipesToRemove = new ArrayList<>();
         for (IRecipe recipe : ForgeRegistries.RECIPES) {
             if (recipe.getRegistryName() != null) {
                 for (IIngredient ingredient : ingredients) {
                     if (ingredient.test(recipe.getRecipeOutput())) {
-                        recipesToRemove.add(recipe.getRegistryName());
+                        ReloadableRegistryManager.removeRegistryEntry(ForgeRegistries.RECIPES, recipe.getRegistryName());
                         break;
                     }
                 }
-            }
-        }
-        if (!recipesToRemove.isEmpty()) {
-            for (ResourceLocation loc : recipesToRemove) {
-                ReloadableRegistryManager.removeRegistryEntry(ForgeRegistries.RECIPES, loc);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/JustEnoughItems.java
```diff
@@ -35,10 +42,43 @@ public void removeAndHide(IIngredient ingredient) {
...
+                return;
+            }
+            JeiPlugin.hideItem(ingredient.getMatchingStacks());
+        }
+        List<ResourceLocation> recipesToRemove = new ArrayList<>();
+        for (IRecipe recipe : ForgeRegistries.RECIPES) {
+            if (recipe.getRegistryName() != null) {
+                for (IIngredient ingredient : ingredients) {
+                    if (ingredient.test(recipe.getRecipeOutput())) {
+                        recipesToRemove.add(recipe.getRegistryName());
```
WaitingIdly, line 61: why have a `recipesToRemove` list instead of simply running `ReloadableRegistryManager.removeRegistryEntry(ForgeRegistries.RECIPES, recipe.getRegistryName());` here?

WaitingIdly (changes requested):
This adds varargs for `removeAndHide`/`yeet`, but is missing varargs for `hide`. Additionally, in places where we use varargs removal we typically have a `Collection<T>` version, to allow either `remove(x,y,z)` or `remove([x, y, z])`.

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/JustEnoughItems.java
```diff
@@ -35,10 +50,64 @@ public void removeAndHide(IIngredient ingredient) {
...
+                        .post();
+                return;
+            }
+            JeiPlugin.hideItem(ingredient.getMatchingStacks());
+        }
+        for (IRecipe recipe : ForgeRegistries.RECIPES) {
+            if (recipe.getRegistryName() != null) {
+                for (IIngredient ingredient : ingredients) {
+                    if (ingredient.test(recipe.getRecipeOutput())) {
+                        ReloadableRegistryManager.removeRegistryEntry(ForgeRegistries.RECIPES, recipe.getRegistryName());
```
brachy84, line 68: Why does it use a different method than the single param method?

brachy84 (changes requested):

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/JustEnoughItems.java
Testure: To avoid multiple iterations of the recipe registry

brachy84 (approved):

PR #71, opened 2023-07-01
add AE2 compat

WaitingIdly:
what this PR adds:
- Add Groovyscript compat for the AE2 Inscriber, AE2 Grinder, allowed Cannon Ammo & damage, P2P Attunement items/mods/capabilities, and allowed Tile Entities moved via Spatial Storage
- a tunnel bracket handler to get AE2 tunnel types
- examples file for the above


examples/appliedenergistics2.groovy
```diff
@@ -0,0 +1,90 @@
+
+// Can be access via either `appliedenergistics2` or `ae2`
+
+// Bracket Handlers
+
+// Get the P2P Tunnel type from the enum. Case insensitive.
+tunnel('me')
```
brachy84, line 7: I wonder if it would be better to add something like this to the mod compat class so it would look like 'mods.ae2.tunnel('me')'

src/main/java/com/cleanroommc/groovyscript/compat/mods/appliedenergistics2/Grinder.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public void removeAll() {
+        for (IGrinderRecipe recipe : AEApi.instance().registries().grinder().getRecipes()) {
+            AEApi.instance().registries().grinder().removeRecipe(recipe);
+            addBackup(recipe);
+        }
+    }
+
+
```
brachy84, line 65: 2 line gap :waaaaahhhh:

examples/appliedenergistics2.groovy
WaitingIdly: at the very least having that as an alternate way to get bracket handlers would be a good idea.

brachy84 (approved):

PR #80, opened 2023-08-01
Apply class transformers to classes loaded by groovy

brachy84:
Since version 4, groovy decompiles the classes itself via asm and is no longer a bunch of reflection magic.
Class formers where not properly (or not at all) applied to classes loaded by groovy. This PR fixes that. 

Meaning groovy will now respect `@SideOnly` annotations. But it does not work in all cases. When a method was removed on server side because it's client annotated and the method is abstract in super class which does not have the annotation, then groovy crashes.


WaitingIdly (approved):
didnt explode the scripts i tested

PR #95, opened 2023-11-21
Lootscript

Wong-Innovations:
Fixes for loot compat not reloading as well as conflicts with other mods.


src/main/java/com/cleanroommc/groovyscript/registry/ReloadableRegistryManager.java
```diff
@@ -68,6 +68,7 @@ public static <T> List<T> unmarkScripted(Class<?> registryClass, @SuppressWarnin
     @ApiStatus.Internal
     public static void init() {
         registryDummies.put(IRecipe.class, DummyRecipe::new);
+        new VanillaModule().init();
```
brachy84, line 71: what
Why not just `VanillaModule.loot.init()` :concern:

Wong-Innovations: Because I figured other VanillaModules added in the future may require some sort of setup, silly goose.

if you prefer it that way I can change it.

brachy84: At least make the method static...

Wong-Innovations: ah like these calls here right? 
image
I already reverted it to the way you'd initially suggested but if you'd like me to make VanillaModule.onReload() and afterScriptRun() static as well let me know. I can bundle that in with these changes.

brachy84: Wtf, i never noticed that

src/main/java/com/cleanroommc/groovyscript/gameobjects/GameObjectHandlers.java
```diff
@@ -205,4 +208,12 @@ public static Result<Material> parseBlockMaterial(String mainArg, Object... args
...
         return material == null ? Result.error() : Result.some(material);
     }
+
+    public static @NotNull Result<NBTTagCompound> parseNBT(String mainArg, Object... args) {
+        try {
+            return Result.some(JsonToNBT.getTagFromJson(mainArg));
+        } catch (NBTException e) {
+            return Result.error("unable to parse provided nbt string");
+        }
+    }
```
brachy84, line 218: I dont think this is a good idea. This requires the nbt to be a string, but its usually a map. I think it would be better to make a method in `GroovyHelper` (its static star imported in all scripts) with a mao argument. There are methods in grs somewhere to convert map to nbt.

src/main/java/com/cleanroommc/groovyscript/compat/loot/GroovyLootFunction.java
```diff
@@ -25,7 +25,10 @@ public GroovyLootFunction(LootCondition[] conditions, Closure<Object> function)
         if (Arrays.equals(function.getParameterTypes(), new Class[]{ItemStack.class, Random.class, LootContext.class})) {
             this.function = function;
         } else {
-            GroovyLog.msg("Error creating custom loot function: function must take the following parameters (net.minecraft.item.ItemStack, java.util.Random, net.minecraft.world.storage.loot.LootContext).").error().post();
+            GroovyLog.msg("Error creating LootFunction:")
+                    .add("function must take the following parameters (net.minecraft.item.ItemStack, java.util.Random, net.minecraft.world.storage.loot.LootContext)")
+                    .error()
+                    .post();
             this.function = null;
```
brachy84, line 32: Enforcing parameter types i something i dont really like. In groovy you can normally just omit the types.

brachy84 (changes requested):

src/main/java/com/cleanroommc/groovyscript/compat/loot/GroovyLootFunction.java
Wong-Innovations: Interesting, I understand why someone would want to omit the types for brevity but I also just figured a very clear error would help save us headaches from end users asking in the discord "why am I getting 'java.util.Random does not have method .setCount(int)' error"

I can revert this and change the docs if this is the consensus however.

src/main/java/com/cleanroommc/groovyscript/compat/loot/Loot.java
```diff
@@ -7,50 +7,44 @@
...
 import net.minecraft.world.storage.loot.conditions.LootCondition;
-import net.minecraft.world.storage.loot.conditions.RandomChance;
-import net.minecraft.world.storage.loot.conditions.RandomChanceWithLooting;
 import org.jetbrains.annotations.ApiStatus;
 
-import java.util.Arrays;
 import java.util.Map;
-import java.util.Random;
 
 public class Loot {
```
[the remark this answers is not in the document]
Wong-Innovations: Good catch, fixed: https://github.com/CleanroomMC/GroovyScript/pull/95/commits/15e26308e1f06518872695892f03adfdb26be60c

src/main/java/com/cleanroommc/groovyscript/gameobjects/GameObjectHandlers.java
Wong-Innovations: Sounds good, I'll give a crack at this implementation.

Wong-Innovations: Allowed for passing a groovy nbt map in: [07e9dd1](https://github.com/CleanroomMC/GroovyScript/pull/95/commits/07e9dd17fad81f2392401bc8a5854cbd0ed45462)

src/main/java/com/cleanroommc/groovyscript/compat/loot/GroovyLootFunction.java
Wong-Innovations: set warnings about types to debug level and made type specification optional: [905b168](https://github.com/CleanroomMC/GroovyScript/pull/95/commits/905b16810d23ca19426a8647bd3ad2112600e09d)

brachy84 (approved):
Im still not happy about the closure param txpes situation, but well figure that out in the future

PR #130, opened 2024-02-25
Add Language Server and vsc companion extension

zznty:
Basic impl of Microsoft LSP protocol for Groovy with a bunch of compatibilities for GroovyScript (dynamic property resolution, blacklist, game object handlers).

### Groovy
- [x] Compiler diagnostics (syntax errors)
- [x] Completions of imports, classes, methods, fields, variables, typed properties
- [x] Quick fix for adding missing import
- [x] Go to definition (only inside referenced groovy classes)
- [x] Go to references (only inside referenced groovy classes)
- [x] Hover signature (with groovydoc support)
- [x] Rename provider
### GroovyScript
- [x] Resolution of implicit imports
- [x] Resolution of globally scoped bindings
- [x] Resolution of statically imported classes
- [x] Resolution of typed properties from `IDynamicGroovyProperty` implementations
- [x] Resolution of `GameObjectHandler` globally scoped methods (only names, due to limited type support in current design state)
- [x] Whitelist/Blacklist support
- [x] MkDoc method description annotations support (limited to methods from `IScriptReloadable` implementations)
- [x] `runConfig.json` support instead of full workspace import


Supernoobv:
we are so back

PR #131, opened 2024-02-26
add Better With Mods compat

WaitingIdly:
what this PR adds:
- compat with `anvilCrafting`, `cauldron`, `crucible`, `kiln`, `millStone`, `saw`, `turntable`, `heat`, `hopper`, and `hopperFilters`
- `anvilCrafting` has all the abilities of a normal crafting table (recipeFunction and recipeAction), which i believe adds novel functionality
- examples for the above
- i18n keys for the above

CraftTweaker, through ModTweaker, has the following explicit compat with Better With Mods that is _not_ added in this PR:
- `bellows`, controlling how far specific items are blown by the Bellows block
- `pulleymanager`, allows adding what blocks can be pulleys
- `buoyancy`, controlling if specific items float or sink in water
- `hcfurnace`, allows setting the time in ticks a given input smelting item takes (only active with the hardmode module enabled)
- `hcmovement`, allows controlling how fast the player walks on specific blocks (only active with the hardmode module enabled)

these werent added because
1. i got tired
2. they already have java methods to set the value and the actual Map used to store the data is a public field
3. they dont have much benefit for being reloadable

relevant classes for anyone who wants access to those, or me if i decide to implement compat
```
betterwithmods.common.registry.BellowsManager
betterwithmods.common.registry.PulleyStructureManager
betterwithmods.module.hardcore.world.HCBuoy
betterwithmods.module.hardcore.crafting.HCFurnace
betterwithmods.module.hardcore.needs.HCMovement
```

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -43,6 +43,13 @@ dependencies {
         runtimeOnly rfg.deobf('curse.maven:chisel-235279:2915375')
     }
 
+    compileOnly rfg.deobf('curse.maven:bwm-suite-246760:3289033')
+    compileOnly rfg.deobf('curse.maven:bwm-core-294335:2624990')
+    if (project.debug_better_with_mods.toBoolean()) {
+        runtimeOnly rfg.deobf('curse.maven:bwm-suite-246760:3289033')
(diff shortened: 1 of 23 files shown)


brachy84 (changes requested):
Same stuff as the other prs

PR #133, opened 2024-02-26
add Alchemistry compat

WaitingIdly:
what this PR adds:
- Adds compat for the `atomizer`, `combiner`, `dissolver`, `electrolyzer`, `evaporator`, `liquifier`.
- i18n wiki keys for the above
- examples file for the above
- `element` game object handler to get the alchemistry element or compound for the given name

something that is not added in this PR:
- crafttweaker, through native compat, adds Create Element and Create Compound helpers. these are not added in this PR. a TODO is added for these helpers.

note: when testing, i suggest also enabling Draconic Evolution for the Creative Power Source (no way to get power otherwise)

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/compat/mods/alchemistry/Atomizer.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/mods/alchemistry/Atomizer.java
@@ -8,10 +8,8 @@ import com.cleanroommc.groovyscript.api.IIngredient;
 import com.cleanroommc.groovyscript.api.documentation.annotations.*;
 import com.cleanroommc.groovyscript.compat.mods.ModSupport;
 import com.cleanroommc.groovyscript.helper.SimpleObjectStream;
-import com.cleanroommc.groovyscript.helper.ingredient.IngredientHelper;
 import com.cleanroommc.groovyscript.helper.recipe.AbstractRecipeBuilder;
 import com.cleanroommc.groovyscript.registry.VirtualizedRegistry;
-import net.minecraft.item.ItemStack;
 import net.minecraftforge.fluids.FluidStack;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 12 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/mods/alchemistry/Atomizer.java
```diff
@@ -0,0 +1,131 @@
...
+    public boolean remove(AtomizerRecipe recipe) {
+        if (ModRecipes.INSTANCE.getAtomizerRecipes().removeIf(r -> r == recipe)) {
+            addBackup(recipe);
+            return true;
+        }
+        return false;
+    }
+
+    @MethodDescription(description = "groovyscript.wiki.removeByOutput", example = @Example(value = "item('alchemistry:compound:7')", commented = true))
+    public boolean removeByOutput(ItemStack output) {
```
brachy84, line 52: Rather use `IIngredient` and `.test()`. This applies to other machines for input and output as well.

src/main/java/com/cleanroommc/groovyscript/compat/mods/alchemistry/Atomizer.java
```diff
@@ -0,0 +1,131 @@
...
+                return true;
+            }
+            return false;
+        });
+    }
+
+    @MethodDescription(description = "groovyscript.wiki.removeByInput")
+    public boolean removeByInput(IIngredient input) {
+        return removeByInput(IngredientHelper.toFluidStack(input));
+    }
```
brachy84, line 76: You can't just use `IIngredient` and then cast to `FluidStack` 💀

src/main/java/com/cleanroommc/groovyscript/compat/mods/alchemistry/Electrolyzer.java
```diff
@@ -0,0 +1,180 @@
...
+                }
+            }
+            return false;
+        });
+    }
+
+    @MethodDescription(description = "groovyscript.wiki.removeByInput")
+    public boolean removeByInput(IIngredient input) {
+        return removeByInput(IngredientHelper.toItemStack(input));
+    }
```
brachy84, line 95: 💀

brachy84 (changes requested):
Please add simple add methods for those who dont like builders

src/main/java/com/cleanroommc/groovyscript/compat/mods/alchemistry/Atomizer.java
WaitingIdly: :skull:

src/main/java/com/cleanroommc/groovyscript/compat/mods/alchemistry/Atomizer.java
WaitingIdly: alright, changed them. there are a few parts of older compat that will also need this change. uh, most really.

brachy84 (approved):

PR #137, opened 2024-03-06
Cache compiled scripts

brachy84:
This pr allows compiled scripts to be cached between reloads and even restarts. This can speed up subsequent reloads significantly.
The Supersymmetry modpack has an average reload (just groovy not restart) time of 70s. With this pr its only 3 - 7s.

## How it works
Groovy compiles scripts into java classes. Before it creates the clas we can obtain the class bytes with a mixin. The class bytes will be immediatly written to a file and out put to `cache/groovy/class.name.clz`. `cache/` is also used by other mods. `.clz` is not a known file ending afaik. Its just class, but clazz and shorter. The class name, bytes, script path, last modified date and inner class names are saved to a map, which is written to `cache/groovy/_index.json`.
On restart the index is read. When a script is about to load we can simply check if it exists in the index and define a class with the bytes. When a script is deleted, the cache (and for all inner classes) is automatically deleted.

## Additional Information
- I added a `/grs deleteClassCache` command wich clears the `cache/groovy/` path.
- I added a `version` property to the index file. This allows us to automatically delete cache in the future, in case we make changes to loading/saving caches wich are not backwards compatible.
- It also works with preprocessors. Now they are only parsed when a script is about to be compiled and saved to the index fle.
- Loaders which are not configured in the run config are no longer being executed.


src/main/java/com/cleanroommc/groovyscript/command/GSCommand.java
```diff
@@ -179,6 +179,10 @@ public GSCommand() {
                                        .appendSibling(GSCommand.getTextForFile("Groovy Log", GroovyLog.get().getLogFilerPath().toString(), new TextComponentString("Click to open GroovyScript log"))));
         }));
 
+        addSubcommand(new SimpleCommand("deleteClassCache", (server, sender, args) -> {
```
WaitingIdly, line 182: "classes" are something distinct from other scripts in GroovyScript, but this clears all cached files. think this should be named `deleteCache`

WaitingIdly (approved):
should preInit scripts be cached? how helpful is caching them?
anyways, ive tested in a few places, and havent gotten any errors related to this change. my comments are just about the part users are going to interact with (the command)

PR #147, opened 2024-03-24
add Nature's Aura compat

WaitingIdly:
changes in this PR:
- Adds `Altar`, `TreeRitual`, `AnimalSpawner`, and `Offering` compat
- adds examples file and i18n wiki keys for the above compat.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/compat/mods/naturesaura/Altar.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/mods/naturesaura/Altar.java
@@ -11,13 +11,12 @@ import com.cleanroommc.groovyscript.registry.VirtualizedRegistry;
 import de.ellpeck.naturesaura.api.NaturesAuraAPI;
 import de.ellpeck.naturesaura.api.recipes.AltarRecipe;
 import net.minecraft.util.ResourceLocation;
-import org.apache.commons.lang3.tuple.Pair;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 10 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/mods/naturesaura/Altar.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Override
+    public void onReload() {
+        removeScripted().forEach(x -> NaturesAuraAPI.ALTAR_RECIPES.remove(x.getKey()));
+        restoreFromBackup().forEach(x -> NaturesAuraAPI.ALTAR_RECIPES.put(x.getKey(), x.getValue()));
+    }
+
+    public void add(AltarRecipe recipe) {
+        if (recipe == null) return;
+        addScripted(Pair.of(recipe.name, recipe));
```
brachy84, line 43: Does it really need to be a pait if the recipe already stores the name?

brachy84 (changes requested):

src/main/java/com/cleanroommc/groovyscript/compat/mods/naturesaura/Altar.java
WaitingIdly: good point. ive made the change here, but Roots compat uses a similar design for almost all of its compat - will need to circle back through and adjust all of those. out of scope for this PR though.

brachy84 (approved):

PR #154, opened 2024-04-02
shorten dependencies and make it cleaner

WaitingIdly:
changes in this PR:
- simply the dependencies file by converting most mods into a map 
- remove duplicate mods
- halve line length
- should make it so there are less merge conflicts and they are easier to resolve


brachy84 (approved):
good stuff

PR #155, opened 2024-04-03
Create AbstractCraftingRecipeBuilder

WaitingIdly:
changes in this PR:
- change Astral Sorcery Starlight Altar to always require 1 starlight, and update example to have 500 starlight
- improve description of `@Property` javadoc, and restrict its use so it always works (can no longer have `@Properties` on FIELD/METHOD, since those are either singleton or need to use `@RecipeBuilderDescription#requirements`)
- fixed Astral Sorcery Starlight Altar documentation generation not quite working correctly, and simplified internal code.
- adds `AbstractCraftingRecipeBuilder` and removes a huge amount of duplicate code relating to that.


brachy84 (approved):
nice

PR #171, opened 2024-06-06
add Immersive Petroleum compat

WaitingIdly:
changes in this PR:
- adds `Distillation` and `Reservoir` compat.
- adds examples file and i18n wiki keys for the above compat.


src/main/java/com/cleanroommc/groovyscript/compat/mods/immersivepetroleum/Distillation.java
```diff
@@ -0,0 +1,171 @@
...
+            return this;
+        }
+
+        @RecipeBuilderMethodDescription
+        public RecipeBuilder chance(Collection<Float> chances) {
+            for (float chance : chances) {
+                chance(chance);
+            }
+            return this;
+        }
```
brachy84, line 132: Are these chances for specific output? If yes, its not a good idea to do it like this. Instead have a `input(IIngredient, float)` method

brachy84 (changes requested):

brachy84 (approved):

PR #173, opened 2024-06-06
Add ProdigyTech compatibility

Wizzerinus:
wow this is huge. I mean the mod does have a lot of things in it

I don't commonly work with Java so might have made some mistakes, the mod's recipe handling system is ridiculous (wow wee constant checks against ItemStack/OreDictIngredient because the author special cased them - I love that so much), and also some GS integrations are somewhat inconsistent (i.e. not sure what methods should be marked with @GroovyBlacklist). Otherwise it seems to work.

List of machines added: `ExplosionFurnace`, `ExplosionFurnaceAdditives`, `RotaryGrinder`, `MagneticReassembler`, `HeatSawmill`, `OreRefinery`, `Solderer`, `PrimordialisReactor`, `AtomicReshaper`, `ZorraAltar`.


src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
```diff
@@ -0,0 +1,166 @@
...
+    public void removeAll() {
+        // NOTE: can't be rolled back due to ItemMap<T> being protected
+        AtomicReshaperManager.INSTANCE.removeAll();
+    }
+
+    @Property(property = "input", valid = @Comp("1"))
+    @Property(property = "output", valid = @Comp("1"))
+    public static class RecipeBuilder extends AbstractRecipeBuilder<AtomicReshaperManager.AtomicReshaperRecipe> {
+
+        @Property(valid = @Comp(value = "1", type = Comp.Type.GTE))
```
WaitingIdly, line 75: should add `defaultValue = "-1"`

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
```diff
@@ -0,0 +1,166 @@
...
+        private int time = -1;
+
+        @Property(valid = @Comp(value = "1", type = Comp.Type.GTE))
+        private int primordium;
+
+        @Property(valid = @Comp("1"))
+        private List<Object> weightedOutputs;
```
WaitingIdly, line 82: you should move the `new ArrayList<>();` up to here

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    public void add(AtomicReshaperManager.AtomicReshaperRecipe recipe) {
+        AtomicReshaperManager.INSTANCE.addRecipe(recipe);
+        addScripted(recipe);
+    }
+
+    @MethodDescription(priority = 2000, example = @Example(commented = true))
+    public void removeAll() {
+        // NOTE: can't be rolled back due to ItemMap<T> being protected
```
WaitingIdly, line 67: in situations like this, you should create an accessor to access the field, such as we do for Thermal Expansion: https://github.com/CleanroomMC/GroovyScript/blob/master/src/main/java/com/cleanroommc/groovyscript/core/mixin/thermalexpansion/BrewerManagerAccessor.java

it takes a few steps, as you need to do three things:
- add the modid to `LateMixin`
- create file called `mixin.groovyscript.[modid].json` in `resources`
- create the accessor and add it to the above json

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
```diff
@@ -0,0 +1,115 @@
...
+import com.cleanroommc.groovyscript.api.IIngredient;
+import com.cleanroommc.groovyscript.api.documentation.annotations.Example;
+import com.cleanroommc.groovyscript.api.documentation.annotations.MethodDescription;
+import com.cleanroommc.groovyscript.api.documentation.annotations.RegistryDescription;
+import com.cleanroommc.groovyscript.helper.ingredient.OreDictIngredient;
+import com.cleanroommc.groovyscript.registry.VirtualizedRegistry;
+import lykrast.prodigytech.common.recipe.ExplosionFurnaceManager;
+import net.minecraft.item.ItemStack;
+
+class EFAdditiveRecipe {
```
WaitingIdly, line 12: could you move this to a private inner class

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/Solderer.java
```diff
@@ -0,0 +1,160 @@
...
+        public String getErrorMsg() {
+            return "Error adding ProdigyTech Solderer Recipe";
+        }
+
+        @Override
+        public void validate(GroovyLog.Msg msg) {
+            validateItems(msg, 0, 1, 1, 1);
+            validateFluids(msg);
+            msg.add(gold <= 0, "gold must be greater than or equal to 1, yet it was {}", gold);
+            msg.add(pattern == null, "pattern cannot be empty");
```
WaitingIdly, line 130: should check if its empty instead of null, since `IIngredient.EMPTY` exists.
```suggestion
            msg.add(IngredientHelper.isEmpty(pattern), "pattern must not be empty");
```

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/Solderer.java
```diff
@@ -0,0 +1,160 @@
...
+        });
+    }
+
+    @MethodDescription(type = MethodDescription.Type.QUERY)
+    public SimpleObjectStream<SoldererManager.SoldererRecipe> streamRecipes() {
+        return new SimpleObjectStream<>(SoldererManager.RECIPES)
+                .setRemover(this::remove);
+    }
+
+    @Property(property = "input", valid = @Comp("1"))
```
WaitingIdly, line 89: because the actual allowed value is a range, this should be more like
```suggestion
    @Property(property = "input", valid = {@Comp(type = Comp.Type.GTE, value = "0"), @Comp(type = Comp.Type.LTE, value = "1")})
```

examples/preInit/prodigytech.groovy
```diff
@@ -0,0 +1,23 @@
+
+// Auto generated groovyscript example file
```
WaitingIdly, line 2: i dont think this file is autogenerated anymore, comment should probably get removed

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnace.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @MethodDescription(example = @Example("item('prodigytech:ferramic_ingot')"))
+    public void removeByOutput(ItemStack output) {
+        ExplosionFurnaceManager.RECIPES.removeIf(r -> {
+            if (!r.getOutput().isItemEqual(output)) return false;
+            addBackup(r);
+            return true;
+        });
+    }
+
```
WaitingIdly, line 46: can we get a `removeAll` and `streamRecipes`? here

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
```diff
@@ -0,0 +1,115 @@
...
+                for (ItemStack it : this.input.getMatchingStacks()) {
+                    ExplosionFurnaceManager.removeDampener(it);
+                }
+            }
+        }
+    }
+}
+
+@RegistryDescription
+public class ExplosionFurnaceAdditives extends VirtualizedRegistry<EFAdditiveRecipe> {
```
WaitingIdly, line 65: can we get a `removeAll` for both explosives and dampeners?

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/PrimordialisReactor.java
```diff
@@ -0,0 +1,51 @@
...
+    public void addRecipe(IIngredient x) {
+        addScripted(x);
+        addRecipeBase(x);
+    }
+
+    @MethodDescription(example = @Example("ore('sugarcane')"))
+    public void removeRecipe(IIngredient x) {
+        addBackup(x);
+        removeRecipeBase(x);
+    }
```
WaitingIdly, line 50: can we get a `removeAll` and `streamRecipes` here?

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
```diff
@@ -0,0 +1,166 @@
...
+            }
+        }
+
+        private Object[] getRecipeOutput() {
+            if (!output.isEmpty()) {
+                return new ItemStack[]{output.get(0)};
+            } else {
+                return weightedOutputs.toArray();
+            }
+        }
```
WaitingIdly, line 144: the way `weightedOutputs` works is nasty. i would suggest replacing the `weightedOutputs` Object List with a `chance` Integer List. then, do this
```suggestion
        private Object[] getRecipeOutput() {
            List<Object> target = new ArrayList<>();
            for (int i = 0; i < output.size(); i++) {
                target.add(output.get(i));
                target.add(chance.get(i));
            }
            return target.toArray();
        }
```
this would also significantly simplify validation, since now you just have to allow `>=1` outputs and check that `chance.size()` and `output.size()` are equal.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
```diff
@@ -0,0 +1,166 @@
...
+        @Example(".input(item('minecraft:gold_block')).addOutput(item('minecraft:diamond_block'), 10).addOutput(item('minecraft:carrot'), 3).primordium(7)")
+    })
+    public AtomicReshaper.RecipeBuilder recipeBuilder() {
+        return new AtomicReshaper.RecipeBuilder();
+    }
+
+    @Override
+    public void onReload() {
+        removeScripted().forEach(recipe -> {
+            if (recipe.isOreRecipe())
```
WaitingIdly, line 31: in situations where theres a line break needed between the if statement and the body, could you put it in curly braces `{}`

WaitingIdly (changes requested):

src/main/resources/assets/groovyscript/lang/en_us.lang
```diff
@@ -1779,6 +1779,27 @@ groovyscript.wiki.woot.stygian_iron_anvil.base.value=Sets the itemstack used for
 groovyscript.wiki.woot.stygian_iron_anvil.preserveBase.value=Sets if the base is used as a catalyst and not consumed
 groovyscript.wiki.woot.stygian_iron_anvil.removeByBase=Removes all recipes that match the given base item
 
+# ProdigyTech
```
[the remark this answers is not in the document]
Wizzerinus: Yeah I just put it above Pyrotech, I guess I'll just move both.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ZorraAltar.java
```diff
@@ -0,0 +1,84 @@
...
+import it.unimi.dsi.fastutil.objects.Object2ObjectOpenHashMap;
+
+import lykrast.prodigytech.common.recipe.ZorraAltarManager;
+import lykrast.prodigytech.common.util.Config;
+import net.minecraft.enchantment.Enchantment;
+
+import java.util.Map;
+
+
+class ZorraRecipeData {
```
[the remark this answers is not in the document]
Wizzerinus: I am aware of records but aren't they only Java 14+? Does GS do some wild magic to transform its records to classic-style classes when ran on Java 8?

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
Wizzerinus: Yea this was just me testing problems with that bizarre crash

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ZorraAltar.java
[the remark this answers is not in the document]
brachy84: We use jabel. It allows to use java 17 syntax on java 8.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
Wizzerinus: would have to be public because
`'com.cleanroommc.groovyscript.compat.mods.prodigytech.ExplosionFurnaceAdditives.EFAdditiveRecipe' has private access in 'com.cleanroommc.groovyscript.compat.mods.prodigytech.ExplosionFurnaceAdditives'`
but otherwise yes I'll do it

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
Wizzerinus: So here, I have a very bizarre issue I am not sure if it is fixable,

so you see, the Explosive's ItemStack is only exposed as a private final property. As far as I know this means it is getting optimized by the JVM in the way where even mixins can't access it. This means `removeAllExplosives()` and such won't be possible to roll back.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/PrimordialisReactor.java
Wizzerinus: `removeAll` done, for `streamRecipes` I would like to use the `Stream` instead of `SimpleObjectStream` but it lacks `setRemover` method, should I just not set any remover?

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
Wizzerinus: notably there is a separate `output("one guaranteed output")` and also `output("my output", weight)`, but i guess I'll do this regardless, I do agree this implementation is nasty

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/PrimordialisReactor.java
brachy84: Use `SimpleObjectStream`. Its whole purpose is to be used in `streamRecipes`, because of its remove ability.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
Wizzerinus: to make sure, the project's braces style goes like this, right?
```java
if (x) return y;
if (z_but_with_long_name_so_it_doesnt_fit_on_one_line) {
    return t_but_with_long_name_so_it_doesnt_fit_on_one_line;
)
```

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/PrimordialisReactor.java
Wizzerinus: the problem is there is no `List<IIngredient>`. ProdigyTech exposes `List<ItemStack>` and `List<String>` (for oredicts), and `SimpleObjectStream` does not have `map`. Unless `SimpleObjectStream -> Stream` conversion works or something.

brachy84: What? You just give it a list of recipes

Wizzerinus: PrimordialisReactor does not have recipes in the usual sense, it's just items that it converts into certain amount of primordium, and you can't set the amount yourself (the reactor manages it). So each of its recipes is either an ItemStack or a String.

Wizzerinus: ok I think I figured out what you wanted me to do

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
Wizzerinus: It turns out I didn't actually have to do this and went through all this for nothing 😭 oh well

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
```diff
@@ -0,0 +1,179 @@
...
+
+        @Override
+        public void validate(GroovyLog.Msg msg) {
+            validateItems(msg, 1, 1, 1, Integer.MAX_VALUE);
+            validateFluids(msg);
+
+            // I think this check is not possible to fail at all but still, adding it for consistency
+            msg.add(output.size() != outputWeights.size(), "Outputs and output weights must be the same size!");
+
+            msg.add(!errorMessage.isEmpty(), "Weighted outputs error: {}", errorMessage);
```
WaitingIdly, line 140:
```suggestion
            msg.add(outputWeights.stream().anyMatch(x -> x <= 0), "all weighted outputs must be greater than 0, yet they were {}", outputWeights);
```

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnace.java
```diff
@@ -0,0 +1,121 @@
...
+        return new ExplosionFurnace.RecipeBuilder();
+    }
+
+    @Override
+    public void onReload() {
+        removeScripted().forEach(this::remove);
+        restoreFromBackup().forEach(ExplosionFurnaceManager::addRecipe);
+    }
+
+    private boolean remove(ExplosionFurnaceManager.ExplosionFurnaceRecipe recipe) {
```
WaitingIdly, line 30: this is used for both `streamRecipes` and `removeScripted().forEach` call this method. this is an issue, since `streamRecipes` should put the recipes it removes in the backups list, and `removeScripted().forEach` should definitely not do that.
i would suggest changing the method to include `addBackup`, and just copying the code into the `forEach` on `removeScripted()`

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnace.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        @Override
+        @RecipeBuilderRegistrationMethod
+        public @Nullable ExplosionFurnaceManager.ExplosionFurnaceRecipe register() {
+            if (!validate()) return null;
+            ExplosionFurnaceManager.ExplosionFurnaceRecipe recipe = null;
+            IIngredient inputItem = input.get(0);
+            // We do not do the OreDict check like in other places as it adds far too much code bloat
+            if (craftPerReagent == 0) {
```
WaitingIdly, line 104: i think you should change this logic to be based on if `input.size() == 2`, and then change the defaultValue of `craftPerReagent` to `1`. this would allow you to simplify some of the logic used.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
```diff
@@ -0,0 +1,130 @@
...
+            remove(recipe);
+            return true;
+        }
+        return false;
+    }
+
+    @MethodDescription(example = @Example(priority = 2000, commented = true))
+    public void removeAllDampeners() {
+        ExplosionFurnaceManager.DAMPENERS.getAllContent().forEach(r ->
+            addBackup(new EFAdditiveRecipe(true, new ItemsIngredient(r.getMatchingStacks()), r.getDampening())));
```
WaitingIdly, line 75: this is removing dampeners. should the `isExplosive` boolean be set to `true` here? interacting with dampeners in other locations in this file indicates it should be `false`.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandler.java
```diff
@@ -0,0 +1,109 @@
...
+
+    SimpleRecipeHandler(String name, SimpleRecipeManager instance) {
+        super(name, instance);
+    }
+
+    @Property(property = "input", valid = @Comp("1"))
+    @Property(property = "output", valid = @Comp("1"))
+    public class RecipeBuilder extends AbstractRecipeBuilder<SimpleRecipe> {
+        @Property(valid = @Comp(value = "1", type = Comp.Type.GTE))
+        private int time = -1;
```
WaitingIdly, line 33: why not set it to `getDefaultTime()` by default?

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandler.java
```diff
@@ -0,0 +1,109 @@
...
+                    addRecipe(recipe);
+                }
+            }
+            return recipe;
+        }
+    }
+}
+
+@RegistryDescription
+class RotaryGrinder extends SimpleRecipeHandler {
```
WaitingIdly, line 76: could you move these into public inner classes of `SimpleRecipeHandler`? there should only be one main class per file.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandler.java
```diff
@@ -0,0 +1,109 @@
...
+            validateItems(msg, 1, 1, 1, 1);
+            validateFluids(msg);
+            if (time == -1) {
+                time = getDefaultTime();
+            }
+        }
+
+        @Override
+        public String getErrorMsg() {
+            return String.format("Error adding ProdigyTech %s Recipe", name);
```
WaitingIdly, line 52: pretty sure this needs `SimpleRecipeHandler.this.name` or it will be an empty string (theres a `name` field in `AbstractRecipeBuilder`)

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandlerAbstract.java
```diff
@@ -0,0 +1,82 @@
...
+        this.name = name;
+    }
+
+    protected abstract int getDefaultTime();
+
+    @Override
+    public void onReload() {
+        removeScripted().forEach(x -> instance.removeRecipe(x.getInput()));
+        restoreFromBackup().forEach(r -> {
+            System.out.println("reloading recipe: " + name);
```
WaitingIdly, line 28: hi

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandlerSecondaryOutput.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Property(property = "input", valid = @Comp("1"))
+    @Property(property = "output", valid = {@Comp(value = "1", type = Comp.Type.GTE), @Comp(value = "2", type = Comp.Type.LTE)})
+    public class RecipeBuilder extends AbstractRecipeBuilder<SimpleRecipe> {
+        @Property(valid = @Comp(value = "1", type = Comp.Type.GTE))
+        private int time = -1;
+
+        @Property(valid = @Comp(value = "1", type = Comp.Type.LTE))
+        private float secondaryChance = 1.0f;
```
WaitingIdly, line 35: could you add this as a `defaultValue = "1.0f"`

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandlerSecondaryOutput.java
```diff
@@ -0,0 +1,119 @@
...
+            validateFluids(msg);
+            if (time == -1) {
+                time = getDefaultTime();
+            }
+            msg.add(secondaryChance > 1.0f || secondaryChance < 0.0f, "secondary output has to be between 0 and 1, got {}", secondaryChance);
+        }
+
+        @Override
+        public String getErrorMsg() {
+            return String.format("Error adding ProdigyTech %s Recipe", name);
```
WaitingIdly, line 61: similar name story here, `SimpleRecipeHandlerSecondaryOutput.this.name`

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandlerSecondaryOutput.java
```diff
@@ -0,0 +1,119 @@
...
+                    addRecipe(recipe);
+                }
+            }
+            return recipe;
+        }
+    }
+}
+
+@RegistryDescription
+class HeatSawmill extends SimpleRecipeHandlerSecondaryOutput {
```
WaitingIdly, line 86: could you move this to be a public inner class? same logic as in the other location.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/Solderer.java
```diff
@@ -0,0 +1,160 @@
...
+        addBackup(recipe);
+        return SoldererManager.removeRecipe(recipe.getPattern(), recipe.getAdditive(), 9999) != null;
+    }
+
+    @MethodDescription(example = @Example("item('prodigytech:pattern_circuit_refined')"))
+    public boolean removeByPattern(IIngredient pattern) {
+        SoldererManager.SoldererRecipe recipe = SoldererManager.RECIPES.stream()
+                .filter(r -> pattern.test(r.getPattern()))
+                .findFirst().orElse(null);
+        return remove(recipe);
```
WaitingIdly, line 49: i dont see the need to do this, you should be able to run this directly
```suggestion
        return SoldererManager.RECIPES.removeIf(r -> {
            if (pattern.test(r.getPattern())) {
                addBackup(r);
                return true;
            }
            return false;
        });
```

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/Solderer.java
```diff
@@ -0,0 +1,160 @@
...
+        @Override
+        public void validate(GroovyLog.Msg msg) {
+            validateItems(msg, 0, 1, 1, 1);
+            validateFluids(msg);
+            msg.add(gold <= 0, "gold must be greater than or equal to 1, yet it was {}", gold);
+            msg.add(pattern.isEmpty(), "pattern cannot be empty");
+            int capacity = Config.soldererMaxGold;
+            msg.add(gold > capacity, "gold must be less than or equal to the Solderer's capacity {}, yet it was {}", capacity, gold);
+            if (time <= 0) {
+                time = Config.soldererProcessTime;
```
WaitingIdly, line 134: why not set the default value of the field to `Config.soldererProcessTime`?

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ZorraAltar.java
```diff
@@ -0,0 +1,75 @@
+package com.cleanroommc.groovyscript.compat.mods.prodigytech;
+import com.cleanroommc.groovyscript.api.GroovyBlacklist;
```
WaitingIdly, line 2: there should be a linebreak between package and the import block.

src/main/resources/assets/groovyscript/lang/en_us.lang
```diff
@@ -1189,6 +1189,116 @@ groovyscript.wiki.naturesaura.spawning.entity.value=Sets the entity spawned
 groovyscript.wiki.naturesaura.spawning.removeByEntity=Removes all Altar of Birthing recipes that summon the given entity
 groovyscript.wiki.naturesaura.spawning.removeByName=Removes the Altar of Birthing recipe with the given name
 
+# ProdigyTech
+groovyscript.wiki.prodigytech.atomic_reshaper.title=Atomic Reshaper
+groovyscript.wiki.prodigytech.atomic_reshaper.description=Uses hot air and Primordium to convert items, can have a weighted random based output
+groovyscript.wiki.prodigytech.atomic_reshaper.time.value=The time needed to perform the recipe, in ticks. By default, takes the value from the ProdigyTech's config
+groovyscript.wiki.prodigytech.atomic_reshaper.primordium.value=The amount of Primordium units consumed by this recipe. 1 Primordium equals 100 Primordium units
+groovyscript.wiki.prodigytech.explosion_furnace.title=Explosion Furnace
```
WaitingIdly, line 1197: minor thing, but could you put an additional linbreak between the registries? i think it improves readability

src/main/resources/assets/groovyscript/lang/en_us.lang
```diff
@@ -1189,6 +1189,116 @@ groovyscript.wiki.naturesaura.spawning.entity.value=Sets the entity spawned
 groovyscript.wiki.naturesaura.spawning.removeByEntity=Removes all Altar of Birthing recipes that summon the given entity
 groovyscript.wiki.naturesaura.spawning.removeByName=Removes the Altar of Birthing recipe with the given name
 
+# ProdigyTech
+groovyscript.wiki.prodigytech.atomic_reshaper.title=Atomic Reshaper
+groovyscript.wiki.prodigytech.atomic_reshaper.description=Uses hot air and Primordium to convert items, can have a weighted random based output
```
WaitingIdly, line 1194: could you put a period at the end of the `description` lang keys? its not included by default, and the sentence should have one. honestly, should probably automatically insert a period at the end, its weird and the only lang key that this applies to.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
```diff
@@ -0,0 +1,130 @@
...
+import com.cleanroommc.groovyscript.api.documentation.annotations.Example;
+import com.cleanroommc.groovyscript.api.documentation.annotations.MethodDescription;
+import com.cleanroommc.groovyscript.api.documentation.annotations.RegistryDescription;
+import com.cleanroommc.groovyscript.helper.ingredient.ItemsIngredient;
+import com.cleanroommc.groovyscript.helper.ingredient.OreDictIngredient;
+import com.cleanroommc.groovyscript.registry.VirtualizedRegistry;
+import lykrast.prodigytech.common.recipe.ExplosionFurnaceManager;
+import net.minecraft.item.ItemStack;
+
+@RegistryDescription
```
WaitingIdly, line 13: some of the words that are used for generating the wiki use the term "recipes". this doesnt really fit here, and theres a way to change that to "entries" via `category = RegistryDescription.Category.ENTRIES`

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.Collection;
+import java.util.List;
+
+@RegistryDescription
+public class AtomicReshaper extends VirtualizedRegistry<AtomicReshaperManager.AtomicReshaperRecipe> {
+    @RecipeBuilderDescription(example = {
+        @Example(".input(item('minecraft:gold_ingot')).output(item('minecraft:emerald_block')).primordium(10).time(50)"),
+        @Example(".input(item('minecraft:gold_block')).output(item('minecraft:diamond_block'), 10).output(item('minecraft:carrot'), 3).primordium(7)")
+    })
+    public AtomicReshaper.RecipeBuilder recipeBuilder() {
```
WaitingIdly, line 25: since `RecipeBuilder` is the only thing called that accessible in this file, theres no need to include `AtomicReshaper.` before it, in any of the places its used here. this also applies to the other files. not a big deal, and honestly doesnt need to get changed, but thought i should note it. 
why more word when few word do trick?

src/main/resources/assets/groovyscript/lang/en_us.lang
```diff
@@ -1189,6 +1189,116 @@ groovyscript.wiki.naturesaura.spawning.entity.value=Sets the entity spawned
...
+groovyscript.wiki.prodigytech.atomic_reshaper.time.value=The time needed to perform the recipe, in ticks. By default, takes the value from the ProdigyTech's config
```
WaitingIdly, line 1195: this lang key will generate the following line on the wiki:
```
- `int`. The time needed to perform the recipe, in ticks. By default, takes the value from the ProdigyTech's config. Requires greater than or equal to 1. (Default `-1`).
```
the `defaultValue` prop in the `@Property` annotation is supposed to take the role of providing an identical way to identify what the default value is, and so it essentially duplicates this information, which is confusing.
i would suggest implementing the fields and lang keys similar to how it is done for Extended Crafting:

https://github.com/CleanroomMC/GroovyScript/blob/master/src/main/java/com/cleanroommc/groovyscript/compat/mods/extendedcrafting/CombinationCrafting.java#L119
https://cleanroommc.com/groovy-script/mods/extendedcrafting/combination_crafting#recipe-builder

as this generates a line that looks like this:
```
- `int`. Sets the maximum amount of RF consumed per tick until the entire cost is paid. Requires greater than or equal to 0. (Default `ModConfig.confCraftingCoreRFRate`).
```

src/main/resources/assets/groovyscript/lang/en_us.lang
```diff
@@ -1189,6 +1189,116 @@ groovyscript.wiki.naturesaura.spawning.entity.value=Sets the entity spawned
...
+groovyscript.wiki.prodigytech.atomic_reshaper.primordium.value=The amount of Primordium units consumed by this recipe. 1 Primordium equals 100 Primordium units
+groovyscript.wiki.prodigytech.explosion_furnace.title=Explosion Furnace
+groovyscript.wiki.prodigytech.explosion_furnace.description=Uses an explosive, a dampener, and an optional reagent to convert items, the power values of the recipe, explosive, and dampener should be close to avoid an efficiency loss
+groovyscript.wiki.prodigytech.explosion_furnace.power.value=The power used by this recipe
```
WaitingIdly, line 1199: since this is used for describing the method, not the field, i think the format of these should be `Sets the ...` instead of `The ...`

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    public void add(AtomicReshaperManager.AtomicReshaperRecipe recipe) {
+        AtomicReshaperManager.INSTANCE.addRecipe(recipe);
+        addScripted(recipe);
+    }
+
+    @MethodDescription(priority = 2000, example = @Example(commented = true))
+    public void removeAll() {
+        AtomicReshaperManager.INSTANCE.getAllRecipes().forEach(this::addBackup);
```
WaitingIdly, line 68: can we get a `streamRecipes` for this

WaitingIdly (changes requested):

examples/preInit/prodigytech.groovy
```diff
@@ -0,0 +1,21 @@
...
+
+import lykrast.prodigytech.common.item.IZorrasteelEquipment
+import lykrast.prodigytech.common.recipe.ZorraAltarManager
+
+if (!isLoaded('prodigytech')) return
+println 'mod \'prodigytech\' detected, running script'
+
+// Create an item at the location 'placeholdername:prodigy_stick' enchantable in the Zorra Altar
+// Note: due to the PT's implementation it is difficult to make other mod's items enchantable
+// This merely registers the item, the pre-init script adds the specific enchantments
```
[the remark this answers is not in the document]
Wizzerinus: typo

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/AtomicReshaper.java
Wizzerinus: oh thats way better, that was kinda nasty to do with the old `Object[]` version

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
Wizzerinus: typo

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandler.java
Wizzerinus: I should do it, it caused mental headaches for me when writing it since I mainly use Python and the semantics of that would be meaningfully different

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandler.java
Wizzerinus: oh no...

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ZorraAltar.java
Wizzerinus: I think when I was writing that file Pycharm crashed and removed all newlines from the file. Will fix.

src/main/resources/assets/groovyscript/lang/en_us.lang
Wizzerinus: ty for the tip, will do

Wizzerinus:
all of these should be fixed now

thanks for being patient with a person not very experienced with java ☺️

src/main/java/com/cleanroommc/groovyscript/documentation/Documentation.java
```diff
@@ -31,7 +31,7 @@ public class Documentation {
 
     public static final IFormat DEFAULT_FORMAT = OutputFormat.VITEPRESS;
 
-    private static final boolean GENERATE_EXAMPLES = false;
+    private static final boolean GENERATE_EXAMPLES = true;
```
WaitingIdly, line 34: should be left as false.

src/main/resources/assets/groovyscript/lang/en_us.lang
```diff
@@ -1189,6 +1189,125 @@ groovyscript.wiki.naturesaura.spawning.entity.value=Sets the entity spawned
...
+groovyscript.wiki.prodigytech.magnetic_reassembler.time.value=The time needed to perform the recipe, in ticks. Will run faster than this number when heated over 125 C.
+
+groovyscript.wiki.prodigytech.ore_refinery.title=Ore Refinery
+groovyscript.wiki.prodigytech.ore_refinery.description=Ore processing machine with 1 input, 2 outputs and an optional chance for the 2nd output.
+groovyscript.wiki.prodigytech.ore_refinery.time.value=The time needed to perform the recipe, in ticks. Will run faster than this number when heated over 125 C.
+groovyscript.wiki.prodigytech.ore_refinery.secondaryChance.value=Sets the chance to get the 2nd output.
+
+groovyscript.wiki.prodigytech.primordialis_reactor.title=Primordialis Reactor
+groovyscript.wiki.prodigytech.primordialis_reactor.description=Turns organic matter into Primordium.
+groovyscript.wiki.prodigytech.primordialis_reactor.addRecipe=Adds an item that can be converted.
```
WaitingIdly, line 1228: methods got renamed
```
groovyscript.wiki.prodigytech.primordialis_reactor.add
groovyscript.wiki.prodigytech.primordialis_reactor.remove
```

WaitingIdly (changes requested):

src/main/resources/assets/groovyscript/lang/en_us.lang
Wizzerinus: intended since those arent really recipes. could do `addItem`/ `removeItem`

src/main/java/com/cleanroommc/groovyscript/documentation/Documentation.java
Wizzerinus: I forgor 💀

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
```diff
@@ -0,0 +1,130 @@
...
+    public void addDampener(IIngredient dampener, int power) {
+        EFAdditiveRecipe recipe = new EFAdditiveRecipe(false, dampener, power);
+        add(recipe);
+    }
+
+    @MethodDescription(example = @Example("ore('dustAsh')"))
+    public boolean removeDampener(IIngredient dampener) {
+        for (ItemStack it : dampener.getMatchingStacks()) {
+            ExplosionFurnaceManager.Dampener externalDampener = ExplosionFurnaceManager.findDampener(it);
+            EFAdditiveRecipe recipe = new EFAdditiveRecipe(false, dampener, externalDampener.getDampening());
```
WaitingIdly, line 65: if `ExplosionFurnaceManager.findDampener(it)` doesnt find anything, it returns `null`, which then generates a `NullPointerException` on this line. this also happens to `ExplosionFurnaceManager.findExplosive(it)`.
there should be an `if (externalDampener == null) continue;` to avoid this.

src/main/resources/assets/groovyscript/lang/en_us.lang
WaitingIdly: my point being that when you generate the wiki `groovyscript.wiki.prodigytech.primordialis_reactor.add` is the lang key checked, which doesnt exist. these lang keys should get renamed so they match the method names.

WaitingIdly (changes requested):

WaitingIdly (approved):

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
```diff
@@ -0,0 +1,132 @@
...
+    @MethodDescription(example = @Example("ore('gunpowder')"))
+    public boolean removeExplosive(IIngredient explosive) {
+        for (ItemStack it : explosive.getMatchingStacks()) {
+            ExplosionFurnaceManager.Explosive externalExplosive = ExplosionFurnaceManager.findExplosive(it);
+            if (externalExplosive == null) continue;
+            EFAdditiveRecipe recipe = new EFAdditiveRecipe(true, explosive, externalExplosive.getPower());
+            remove(recipe);
+            return true;
+        }
+        return false;
```
brachy84, line 46: wtf???

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @MethodDescription(example = @Example(priority = 2000, commented = true))
+    public void removeAllDampeners() {
+        ExplosionFurnaceManager.DAMPENERS.getAllContent().forEach(r ->
+            addBackup(new EFAdditiveRecipe(false, new ItemsIngredient(r.getMatchingStacks()), r.getDampening())));
+        ExplosionFurnaceManager.removeAllDampeners();
+    }
+
+    public static class EFAdditiveRecipe {
```
brachy84, line 81: I dont approve how this registry is done. Do not create a custom recipe class. Rather just use the existing explosive class and add a `AbstractReloadableStorage<Dampner>` field. See [this](https://github.com/CleanroomMC/GroovyScript/blob/ab64a4b26791f0848ae6100a001a72a1334c975f/src/main/java/com/cleanroommc/groovyscript/compat/mods/thermalexpansion/machine/Smelter.java#L25) for example

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandler.java
brachy84, line 1: cursed class design

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandlerAbstract.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public boolean removeByInput(IIngredient input) {
+        if (input instanceof OreDictIngredient) {
+            return removeOreRecipe(((OreDictIngredient) input).getOreDict());
+        } else {
+            boolean removed = false;
+            for (ItemStack it : input.getMatchingStacks()) {
+                // || in this order to remove all matching recipes if they exist for some reason
+                removed = removeStackRecipe(it) || removed;
```
brachy84, line 68: removed |= removeStackRecipe(it)

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/Solderer.java
```diff
@@ -0,0 +1,167 @@
...
+        public String getErrorMsg() {
+            return "Error adding ProdigyTech Solderer Recipe";
+        }
+
+        @Override
+        public void validate(GroovyLog.Msg msg) {
+            validateItems(msg, 0, 1, 1, 1);
+            validateFluids(msg);
+            msg.add(gold <= 0, "gold must be greater than or equal to 1, yet it was {}", gold);
+            msg.add(pattern.isEmpty(), "pattern cannot be empty");
```
brachy84, line 139: use `IngredientHelper.isEmpty(pattern)` since it can be null

brachy84 (changes requested):

examples/assets/placeholdername/lang/en_us.lang
```diff
@@ -2,8 +2,21 @@ itemGroup.groovyscript.example_creative_tab=GroovyScript Creative Tab
...
+enchantment.level.11=XI
+enchantment.level.12=XII
+enchantment.level.13=XIII
+enchantment.level.14=XIV
+enchantment.level.15=XV
+enchantment.level.16=XVI
+enchantment.level.17=XVII
+enchantment.level.18=XVIII
+enchantment.level.19=XIX
+enchantment.level.20=XX
```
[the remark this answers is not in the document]
Wizzerinus: This is an example shrugs. I can just rework the example so it doesn't go to 20.

gradlew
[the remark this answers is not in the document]
Wizzerinus: executable bit for linux users. I can revert it but there's no reason to not have it

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandler.java
Wizzerinus: what do you want though lmao, better than having 4 classes with literally the same content

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
Wizzerinus: wasn't aware of the existence of `AbstractReloadableStorage`

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandler.java
Wizzerinus: then say THAT, and not something completely unrelated. Will fix.

brachy84: :trollface:

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
Wizzerinus: Yeah uh, I think this might be the only sensible way to do it.

brachy84: It isnt. if you apply the change the convo below it will become much better

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/Solderer.java
Wizzerinus: pattern is nonnull in the mod code, although I guess I should still account for this in the builder to avoid leg shooting.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
Wizzerinus: Do I not inherit from `VirtualizedRegistry` then? Not sure if I need to do anything if my "registry" doesn't contain recipes in itself

Wizzerinus: Ok so, I looked into the PT's code and it requires an accessor mixin. I feel like a less destructive option is better.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandler.java
Wizzerinus: I can make the sub classes public but I am NOT going to make the builders non static because it would require coding the same thing 4 times (because of single dispatch).

Wizzerinus:
made the adjustments that made sense, not sure what to do with the other 2.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnace.java
```diff
@@ -97,6 +97,7 @@ public void validate(GroovyLog.Msg msg) {
             validateItems(msg, 1, 2, 1, 1);
             validateFluids(msg);
             msg.add(craftPerReagent <= 0, "craftPerReagent should be positive!");
+            msg.add(power <= 0, "power should be positive!");
```
WaitingIdly, line 100: message is "power should be positive", actual check is error if "<= 0". since `0` is positive, the message is incorrect.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/SimpleRecipeHandler.java
brachy84: thats fine then

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
brachy84: Youd still have a `VirtualizedRegistry<Explossive>` with a field `AbstractReloadableStorage<Dampner>`. And there is nothing wrong with accessor mixins.

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnace.java
WaitingIdly: in mathematics yes, in number representation in programming there is a positive and negative zero (`+0`, `-0`). i think you should specify that it must be greater than 0.
https://en.wikipedia.org/wiki/Signed_zero

src/main/java/com/cleanroommc/groovyscript/compat/mods/prodigytech/ExplosionFurnaceAdditives.java
Wizzerinus: made a version of this, unfortunately it is the only sensible version (the other version requires having a custom class too with all the consequences).

brachy84 (approved):
I will probably change 1 or 2 things once its merged.

PR #176, opened 2024-06-08
Add EssentialCraft 4 compatibility

Wizzerinus:
List of machines added: `DemonTrade`, `MagicianTable`, `MagmaticSmeltery`, `MithrilineFurnace`, `RadiatingChamber`, `WindRune`. There are three notes I wanted to add:

* Demon Trade is bugged and will not accept any item with a different NBT, which breaks like 30% of the recipes of the mod itself. While GrS can set the item NBT, this should be probably noted somewhere in the docs, not sure though.
* Magmatic Smeltery currently cannot be reloaded. The machine isn't usually being used to add things at runtime, and I am surprised it works, but then apparently something caches the output of `getSubItems` causing an NPE. Or I am getting trolled by a client-server desync (the most likely). Theoretically I could mixin into the clientside methods of `OreSmeltingRecipe`, but that seems a bit disruptive. I've currently added a check that makes sure to not reload Magmatic Smeltery recipes to avoid the NPE.
* Also on a similar note: Magmatic Smeltery does not have a JEI handler.
* I wanted to add the support for the mod's Modular Guns system, but since it requires having 12 textures per gun material I think it is a bit annoying to set up examples for it, could still do it though if desired.


gradle.properties
```diff
@@ -14,6 +14,7 @@ debug_calculator = false
 debug_chisel = false
 debug_compact_machines = false
 debug_draconic_evolution = false
+debug_ec4 = false
```
WaitingIdly, line 17: could this be the full mod name?

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/DemonTradeManager.java
```diff
@@ -0,0 +1,87 @@
...
+        super(Alias.generateOf("DemonTrade"));
+    }
+
+    @Override
+    public void onReload() {
+        removeScripted().forEach(DemonTrade::removeTrade);
+        restoreFromBackup().forEach(r -> {
+            DemonTrade.TRADES.add(r);
+            if (r.desiredItem.isEmpty()) {
+                DemonTrade.ALL_MOBS.add(r.entityType);
```
WaitingIdly, line 26: i dont believe `ALL_MOBS` is handled properly in all removal cases - consider instead using `afterScriptLoad`, such as in `Precipitator`

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/DemonTradeManager.java
```diff
@@ -0,0 +1,87 @@
...
+import com.cleanroommc.groovyscript.api.documentation.annotations.MethodDescription;
+import com.cleanroommc.groovyscript.api.documentation.annotations.RegistryDescription;
+import com.cleanroommc.groovyscript.helper.Alias;
+import com.cleanroommc.groovyscript.helper.SimpleObjectStream;
+import com.cleanroommc.groovyscript.registry.VirtualizedRegistry;
+import net.minecraft.item.ItemStack;
+import essentialcraft.api.DemonTrade;
+import net.minecraftforge.fml.common.registry.EntityEntry;
+
+@RegistryDescription(category = RegistryDescription.Category.ENTRIES)
```
WaitingIdly, line 14: yeah, i think you should add an admonition here - i think a `DANGER`-type would be good

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MagicianTable.java
```diff
@@ -0,0 +1,88 @@
...
+import com.cleanroommc.groovyscript.api.documentation.annotations.*;
+import com.cleanroommc.groovyscript.helper.SimpleObjectStream;
+import com.cleanroommc.groovyscript.helper.recipe.AbstractRecipeBuilder;
+import com.cleanroommc.groovyscript.registry.VirtualizedRegistry;
+import essentialcraft.api.MagicianTableRecipe;
+import essentialcraft.api.MagicianTableRecipes;
+import net.minecraft.item.crafting.Ingredient;
+import org.jetbrains.annotations.Nullable;
+
+@RegistryDescription(admonition = @Admonition("groovyscript.wiki.essentialcraft.magician_table.note0"))
```
WaitingIdly, line 14: i think this should be a `WARNING`-type admonition

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MithrilineFurnace.java
```diff
@@ -0,0 +1,105 @@
...
+                ItemStack[] options = input.get(0).getMatchingStacks();
+                msg.add(options.length == 0, "must have at least 1 matching stack for the input");
+            }
+        }
+
+        @Override
+        @RecipeBuilderRegistrationMethod
+        public @Nullable MithrilineFurnaceRecipe register() {
+            if (!validate()) return null;
+            int stackSize = input.get(0).getMatchingStacks()[0].getCount();
```
WaitingIdly, line 97:
```suggestion
            int stackSize = input.get(0).getAmount();
```

src/main/resources/assets/groovyscript/lang/en_us.lang
```diff
@@ -796,6 +796,40 @@ groovyscript.wiki.enderio.vat.multipliers2.value=Sets the multiplier applied to
 groovyscript.wiki.enderio.vat.baseMultiplier.value=Sets the base amount of fluid output
 groovyscript.wiki.enderio.vat.energy.value=Sets the energy cost of the recipe
 
+# EssentialCraft
+groovyscript.wiki.essentialcraft.demon_trade.title=Demon Trade
+groovyscript.wiki.essentialcraft.demon_trade.description=Adds an item that can be sold to Demons to obtain Ackronite. Note that each demon that spawns has a random item that it can accept, and will not accept any other item.
+groovyscript.wiki.essentialcraft.demon_trade.add=Adds an item or an entity that can be sold to demons.
```
WaitingIdly, line 802: currently, this creates two periods for the `add` and `remove` on the wiki. should remove the period at the end of the lines for everything except `description`

WaitingIdly (changes requested):
sorry for the delay

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MithrilineFurnace.java
```diff
@@ -0,0 +1,105 @@
...
+            Ingredient inputItem = input.get(0).withAmount(1).toMcIngredient();
+            MithrilineFurnaceRecipe recipe = new MithrilineFurnaceRecipe(inputItem, output.get(0), (float) espe, stackSize);
```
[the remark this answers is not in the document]
Wizzerinus: because it is really an int parameter. Like I could make it a float, but I think it is unintuitive because in game all recipes show up with int costs

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MagmaticSmeltery.java
```diff
@@ -0,0 +1,127 @@
...
+import com.cleanroommc.groovyscript.helper.SimpleObjectStream;
+import com.cleanroommc.groovyscript.helper.recipe.IRecipeBuilder;
+import com.cleanroommc.groovyscript.registry.VirtualizedRegistry;
+import essentialcraft.api.OreSmeltingRecipe;
+import net.minecraftforge.oredict.OreDictionary;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collection;
+
+@RegistryDescription(admonition = {
```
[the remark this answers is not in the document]
Wizzerinus: will do, didn't know about reloadability

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MagicianTable.java
```diff
@@ -0,0 +1,88 @@
...
+    public SimpleObjectStream<MagicianTableRecipe> streamRecipes() {
+        return new SimpleObjectStream<>(MagicianTableRecipes.RECIPES).setRemover(r -> {
+            addBackup(r);
+            return MagicianTableRecipes.RECIPES.remove(r);
+        });
+    }
+
+    @Property(property = "input", valid = {@Comp(value = "1", type = Comp.Type.GTE), @Comp(value = "5", type = Comp.Type.LTE)})
+    @Property(property = "output", valid = @Comp("1"))
+    public class RecipeBuilder extends AbstractRecipeBuilder<MagicianTableRecipe> {
```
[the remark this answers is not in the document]
Wizzerinus: I'm using non-static here to point at the instance of MagicianTable registry to use its methods...?

gradle.properties
Wizzerinus: sure

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/DemonTradeManager.java
Wizzerinus: it's notable that the mod's former maintainer fixed this a few days ago, but will still add it

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/DemonTradeManager.java
Wizzerinus: Is this because of `removeAll` or am I missing anything else? I don't think anything else can remove the entity type

Wizzerinus:
all good! thank you for the review regardless

pushed most fixes, although I don't understand the ALL_MOBS one

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/DemonTradeManager.java
WaitingIdly: removeAll works, but `remove(IIngredient)` and `remove(EntityEntry)` dont

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MagicianTable.java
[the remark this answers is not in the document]
WaitingIdly: use `ModSupport.ELEMENTAL_CRAFT_4.get().magicianTable.[method]` instead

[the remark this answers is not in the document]
Wizzerinus: I am going to update them, but. Is there a particular reason why it should be this way? Is it only for consistency? if I had to imagine there shouldn't be virtually any performance/memory overhead from such a short-lived object

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/DemonTradeManager.java
Wizzerinus:
```java
    public DemonTrade(EntityEntry e) {
        this.desiredItem = ItemStack.EMPTY;
        this.entityType = e;
        ALL_MOBS.add(e);
        TRADES.add(this);
    }

    public static void removeTrade(DemonTrade tra) {
        if (tra.entityType != null) {
            ALL_MOBS.remove(tra.entityType);
        }

        TRADES.remove(tra);
    }
```
So theoretically it should work (although the check for `r.entityType != null` is more paradigmatic). I guess something prevents it from working then, huh. I'll test it further after my thesis defence (so in a few days).

Wizzerinus:
ok this should be all

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MagmaticSmeltery.java
```diff
@@ -0,0 +1,129 @@
...
+    static boolean didReload = false;
+
+    @RecipeBuilderDescription(example = @Example(".input('blockIron').output('ingotGold').factor(3).color(0x0000ff)"))
+    public MagmaticSmeltery.RecipeBuilder recipeBuilder() {
+        return new MagmaticSmeltery.RecipeBuilder();
+    }
+
+    @Override
+    public void onReload() {
+        Collection<OreSmeltingRecipe> scripted = removeScripted();
```
WaitingIdly, line 29: to fix the crash after reloading, do the following:
in `LateMixin`, add `essentialcraft` to the `modMixins` list.
create a mixin file called `mixin.groovyscript.essentialcraft.json` and make sure it points towards `mixin.essentialcraft`.
create the following file in the `mixins/essentialcraft` folder:
```java
package com.cleanroommc.groovyscript.core.mixin.essentialcraft;

import com.llamalad7.mixinextras.injector.wrapoperation.Operation;
import com.llamalad7.mixinextras.injector.wrapoperation.WrapOperation;
import essentialcraft.api.OreSmeltingRecipe;
import org.spongepowered.asm.mixin.Mixin;
import org.spongepowered.asm.mixin.injection.At;

@Mixin(value = OreSmeltingRecipe.class, remap = false)
public abstract class OreSmeltingRecipeMixin {

    @WrapOperation(method = "getColorFromItemStack", at = @At(value = "FIELD", target = "Lessentialcraft/api/OreSmeltingRecipe;color:I"))
    private static int getColorFromItemStack(OreSmeltingRecipe recipe, Operation<Integer> original) {
        if (recipe == null) return 16777215;
        return original.call(recipe);
    }

    @WrapOperation(method = "getLocalizedOreName", at = @At(value = "FIELD", target = "Lessentialcraft/api/OreSmeltingRecipe;oreName:Ljava/lang/String;"))
    private static String getLocalizedOreName(OreSmeltingRecipe recipe, Operation<String> original) {
        if (recipe == null) return "";
        return original.call(recipe);
    }

}
```

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MithrilineFurnace.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    public void onReload() {
+        removeScripted().forEach(MithrilineFurnaceRecipes::removeRecipe);
+        restoreFromBackup().forEach(MithrilineFurnaceRecipes::addRecipe);
+    }
+
+    @MethodDescription(example = @Example("ore('dustGlowstone')"))
+    public boolean removeByInput(IIngredient x) {
+        return MithrilineFurnaceRecipes.RECIPES.removeIf(r -> {
+            if (x.equals(r.input)) {
```
WaitingIdly, line 33: this doesnt work and it always false because `IIngredient` is at no point equal to an `Ingredient`. have to do
`for (var item : r.input.getMatchingStacks()) { x.test(item) ...` instead.

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/RadiatingChamber.java
```diff
@@ -0,0 +1,121 @@
...
+        return RadiatingChamberRecipes.RECIPES.removeIf(r -> {
+            if (x.test(r.getRecipeOutput())) {
+                addBackup(r);
+                return true;
+            }
+            return false;
+        });
+    }
+
+    @MethodDescription(example = @Example(priority = 2000, commented = true))
```
WaitingIdly, line 40: priority should go in `MethodDescription` annotation instead of the `Example` annotation

WaitingIdly (changes requested):
took a bit because i wanted to get the mixin fix for removing smeltery stuff done. which now it is.

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MagmaticSmeltery.java
```diff
@@ -0,0 +1,129 @@
...
+            this.factor = factor;
+            return this;
+        }
+
+        public String getErrorMsg() {
+            return "Error adding Magmatic Smeltery Recipe";
+        }
+
+        public void validate(GroovyLog.Msg msg) {
+            msg.add(!OreDictionary.doesOreNameExist(input), "Unknown input OreDict: {}", input);
```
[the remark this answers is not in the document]
Wizzerinus: Yeah that one is fair

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MagmaticSmeltery.java
Wizzerinus: Holy shit you actually fixed it. Thank you so much

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MithrilineFurnace.java
Wizzerinus: Uh oh, how did it work when I tested it....

Wizzerinus:
all of these resolved, with an extra fix for a magmatic smeltery crash when an invalid ore is inserted

src/main/java/com/cleanroommc/groovyscript/core/mixin/essentialcraft/TileMagmaticSmelterMixin.java
brachy84, line 1: add a comment to the mixins explaining what it does

brachy84 (changes requested):

Wizzerinus:
done

brachy84 (approved):

src/main/java/com/cleanroommc/groovyscript/compat/mods/essentialcraft/MagmaticSmeltery.java
```diff
@@ -0,0 +1,131 @@
...
+import com.cleanroommc.groovyscript.api.documentation.annotations.*;
+import com.cleanroommc.groovyscript.compat.mods.ModSupport;
+import com.cleanroommc.groovyscript.helper.SimpleObjectStream;
+import com.cleanroommc.groovyscript.helper.ingredient.OreDictIngredient;
+import com.cleanroommc.groovyscript.helper.recipe.IRecipeBuilder;
+import com.cleanroommc.groovyscript.registry.VirtualizedRegistry;
+import essentialcraft.api.OreSmeltingRecipe;
+import org.jetbrains.annotations.Nullable;
+
+@RegistryDescription(reloadability = RegistryDescription.Reloadability.DISABLED, admonition={
```
WaitingIdly, line 13: since theres only one admonition, can remove the curly braces around it. should have your IDE highlighting it.

WaitingIdly (approved):

PR #179, opened 2024-06-10
Minor Cleanup of IDE warnings, also small bugfixes

WaitingIdly:
changes in this PR:
- engage in a significant amount of "IDE warning gets autofixed" action
	- weaken overly strong variables
	- less modification of method parameters
	- one line per variable/field declaration
	- remove redundant field creation (eg `null` or `0`, when that is already the default)
	- apply `@Override` annotations where they ought to be
	- make `name` use `super.name` in recipe builders where it can be confused for the `name` in the parent class
	- remove unused imports
	- use pattern variable
	- use improved switch format
	- make finalizable fields final
	- a few more minor things
- make Blood Magic return the recipe in `register` correctly
- store only the recipe in most roots stuff, instead of `Pair<ResourceLocation, Recipe>` (Recipe has the RL, its pointless)
- fix some extended crafting shapeless recipes actually being shaped recipes
- remove some instances of `@MethodDescription(description = "x")` where `x` could be assumed
- add the ability for java arguments to control documentation, and add settings to `gradle.properties` to allow that
- reorder the `gradle.properties` debug stuff and comment it
- fix `mode` being renamed to `compactorMode` not having documentation or examples updated

removes ~480 warnings, which according to my IDE, means only 460 left in the groovyscript folder.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/compat/mods/mekanism/ChemicalOxidizer.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/mods/mekanism/ChemicalOxidizer.java
@@ -34,10 +34,9 @@ public class ChemicalOxidizer extends VirtualizedMekanismRegistry<OxidationRecip
         msg.add(Mekanism.isEmpty(output), () -> "output must not be empty");
         if (msg.postIfNotEmpty()) return null;
 
-        output = output.copy();
         OxidationRecipe recipe1 = null;
         for (ItemStack itemStack : ingredient.getMatchingStacks()) {
-            OxidationRecipe recipe = new OxidationRecipe(itemStack.copy(), output);
+            OxidationRecipe recipe = new OxidationRecipe(itemStack.copy(), output.copy());
             if (recipe1 == null) recipe1 = recipe;
             recipeRegistry.put(recipe);
             addScripted(recipe);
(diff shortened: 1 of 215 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/mods/mekanism/ChemicalOxidizer.java
```diff
@@ -34,10 +34,9 @@ public OxidationRecipe add(IIngredient ingredient, GasStack output) {
         msg.add(Mekanism.isEmpty(output), () -> "output must not be empty");
         if (msg.postIfNotEmpty()) return null;
 
-        output = output.copy();
         OxidationRecipe recipe1 = null;
         for (ItemStack itemStack : ingredient.getMatchingStacks()) {
-            OxidationRecipe recipe = new OxidationRecipe(itemStack.copy(), output);
+            OxidationRecipe recipe = new OxidationRecipe(itemStack.copy(), output.copy());
```
brachy84, line 39: Seems like an unnecessary amount of copying

brachy84 (commented):
Not a huge fan of removing initalisation of fields even if its their default value. It helps make the code more readable and less ambiguous. Also the `1f` -> `1.0f` seems unecessary, but its fine.

brachy84 (approved):

PR #180, opened 2024-06-10
Fix for ls not suggesting items from supers and interaces

### LINKED ISSUE #174: Language Server doesn't suggest parent class completions
Hi!
I noticed that the language server is not aware of fields and methods of the parent class of the one you are trying to get completions for.

Expected behaviour:
immagine

Actual behaviour:
immagine

This issue applies to classes that extend both groovy classes and java ones.

zznty:
Fixes #174

For some reason `ClassNode.getMethods()` returns only declared methods but `ClassNode.getMethods(String name)` does search across all supers and interfaces. With this pr, ls would collect props, fields and methods recusively from all supers and implemented interfaces.


brachy84:
Fields and methods of all super classes should be added. And i dont like the use of streams here. It could just be a list you pass as a param in the collect methods

PR #183, opened 2024-06-22
Add The Aurorian compatibility

Wizzerinus:
The processing machines in the mod are very simple, so should be pretty straightforward

Added machines: `Scrapper`, `MoonlightForge`


WaitingIdly (changes requested):

src/main/java/com/cleanroommc/groovyscript/compat/mods/theaurorian/MoonlightForge.java
```diff
@@ -0,0 +1,91 @@
...
+            if (output.test(r.getOutput())) {
+                addBackup(r);
+                return true;
+            }
+            return false;
+        });
+    }
+
+    @MethodDescription(priority = 2000, example = @Example(commented = true))
+    public void removeAll() {
```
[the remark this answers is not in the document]
Wizzerinus: will do (same with the other one)

Wizzerinus:
probably fixed these comments

src/main/java/com/cleanroommc/groovyscript/compat/mods/theaurorian/MoonlightForge.java
```diff
@@ -0,0 +1,104 @@
...
+        MoonlightForgeRecipeHandler.allRecipes.forEach(this::addBackup);
+        MoonlightForgeRecipeHandler.allRecipes.clear();
+    }
+
+    @MethodDescription(type = MethodDescription.Type.QUERY)
+    public SimpleObjectStream<MoonlightForgeRecipe> streamRecipes() {
+        return new SimpleObjectStream<>(MoonlightForgeRecipeHandler.allRecipes).setRemover(r ->
+            MoonlightForgeRecipeHandler.allRecipes.removeIf(u -> {
```
WaitingIdly, line 65: could this code be moved to a `remove` function?

Wizzerinus: I suppose `removeByInput`/`removeByOutput` don't have to be modified to use that method? Seems like either an unnecessary performance bottleneck or an unnecessary bunch of type casts

WaitingIdly: nope, just prefer having a basic `public boolean remove(T)` method for `streamRecipes` to work with, and thats whats done in most registries using `streamRecipes`

Wizzerinus: should work

WaitingIdly (approved):

src/main/java/com/cleanroommc/groovyscript/compat/mods/theaurorian/MoonlightForge.java
```diff
@@ -0,0 +1,106 @@
...
+    @MethodDescription(example = @Example("item('theaurorian:aurorianiteingot')"))
+    public boolean removeByInput(IIngredient output) {
+        return MoonlightForgeRecipeHandler.allRecipes.removeIf(r -> {
+            if (output.test(r.getInput1()) || output.test(r.getInput2())) {
+                addBackup(r);
+                return true;
+            }
+            return false;
+        });
+    }
```
brachy84, line 53: I feel like this should be
```java
public boolean removeByInput(IIngredient input1, IIngredient input2) {
    return MoonlightForgeRecipeHandler.allRecipes.removeIf(r -> {
        if (input1.test(r.getInput1()) && input2.test(r.getInput2())) {
            addBackup(r);
            return true;
        }
        return false;
    });
}
```

src/main/java/com/cleanroommc/groovyscript/compat/mods/theaurorian/MoonlightForge.java
```diff
@@ -0,0 +1,106 @@
...
+        @RecipeBuilderRegistrationMethod
+        public @Nullable MoonlightForgeRecipe register() {
+            if (!validate()) return null;
+            MoonlightForgeRecipe recipe = null;
+            for (ItemStack input1 : input.get(0).getMatchingStacks()) {
+                for (ItemStack input2 : input.get(1).getMatchingStacks()) {
+                    recipe = new MoonlightForgeRecipe(input1, input2, output.get(0));
+                    ModSupport.THE_AURORIAN.get().moonlightForge.add(recipe);
+                }
+            }
```
brachy84, line 102: I really hate this. Might be better to force item stacks

brachy84 (changes requested):

src/main/java/com/cleanroommc/groovyscript/compat/mods/theaurorian/MoonlightForge.java
Wizzerinus: I mean this is not ideal but I don't see a reason why not?

Wizzerinus:
fixed the former, don't think latter is necessary personally? can still implement that if needed, but just like, why

src/main/java/com/cleanroommc/groovyscript/compat/mods/theaurorian/MoonlightForge.java
```diff
@@ -41,10 +41,10 @@ public boolean remove(MoonlightForgeRecipe recipe) {
         });
     }
 
-    @MethodDescription(example = @Example("item('theaurorian:aurorianiteingot')"))
-    public boolean removeByInput(IIngredient output) {
+    @MethodDescription(example = @Example("item('theaurorian:moonstonesword'), item('theaurorian:aurorianiteingot')"))
+    public boolean removeByInput(IIngredient input, IIngredient catalyst) {
```
WaitingIdly, line 45: should give this a specific lang key, currently it will be using the general `groovyscript.wiki.removeByInput`. theres technically a `groovyscript.wiki.removeByInputAndCatalyst` thats used, but its used all of one place, so i would suggest using a custom lang key description here.

src/main/java/com/cleanroommc/groovyscript/compat/mods/theaurorian/MoonlightForge.java
brachy84: If both inputs are oredicts with 10 entries each (which is rare, but can happen), then it will add 100 recipes. But its probably fine most of the time.

Wizzerinus: Well that's kind of on the modpack dev that they did this? Not to mention if they don't use unidict then that's just something they might have to do anyway

src/main/java/com/cleanroommc/groovyscript/compat/mods/theaurorian/MoonlightForge.java
Wizzerinus: done

WaitingIdly: actually, how viable would doing this be?
```java
    @MethodDescription(example = @Example("item('theaurorian:moonstonesword')")
    public boolean removeByInput(IIngredient input) {
        return MoonlightForgeRecipeHandler.allRecipes.removeIf(r -> {
            if (input.test(r.getInput1()) || input.test(r.getInput2())) {
                addBackup(r);
                return true;
            }
            return false;
        });
```

brachy84: Thats how it was before? I think if it has to inputs you should supply 2 inputs to match for.

Wizzerinus: That's literally what it was before my changes in the last 2 days. I can roll back if needed.

brachy84 (approved):

Wizzerinus:
Accident

PR #187, opened 2024-07-05
Change from mixin GuiScreen to listen to ClientChatEvent

### LINKED ISSUE #78: 0.5.2 crashes with RandomPatches
This doesn't happen with 0.4.6

WaitingIdly:
changes in this PR:
- changes the copy code from being a mixin on `GuiScreen` to being event-based on `ClientChatEvent`.

this is because Random Patches modifies `GuiScreen` via transformer, causes copying not work when the relevant RP option is enabled. tested it locally against Random Patches and this now copies. 

PR changed, the following statement is now implemented:
however, might break on some other unknown mod that transforms or overrides GuiChat? so it might be worth just making a command or using `ClickEvent.Action.RUN_COMMAND` and listening to `ClientChatEvent` to copy it there (so theres no need for a command).

resolves #78


brachy84:
`ClientChatEvent` seems to be perfect for this. I didnt know it existed

brachy84 (approved):

PR #202, opened 2024-07-17
Create recipes using a RecipeBuilder in more situations

WaitingIdly:
changes in this PR:
- add more ability to add Aspects to Thaumcraft Infusion and Crucible Recipe Builders.
- make the Thaumcraft Crucible Recipe Builder create the `CrucibleRecipe` in the register method, instead of calling a different method. 
- add a lens field and validation to Actually Additions Atomic Reconstructor.
- convert (most) cases of adding recipes via shorthand to refer to the relevant Recipe Builder. the benefits of this include:
	- improved error logging due to using the more robust validation methods in every situation
	- reduced code duplication
	- revealing several errors with some of the current add methods
- fixes several add methods using Minecraft's Ingredient instead of GroovyScript's IIngredient or otherwise using types that are harder for GroovyScript to obtain.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/compat/mods/botania/ElvenTrade.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/mods/botania/ElvenTrade.java
@@ -14,6 +14,7 @@ import vazkii.botania.api.BotaniaAPI;
 import vazkii.botania.api.recipe.RecipeElvenTrade;
 
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 
 @RegistryDescription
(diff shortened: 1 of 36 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/mods/botania/ElvenTrade.java
```diff
@@ -38,14 +38,12 @@ protected Object[] convertIngredients(IIngredient[] inputs) {
...
-        add(recipe);
-        return recipe;
+        return recipeBuilder().input(inputs).output(outputs).register();
     }
 
     @MethodDescription(type = MethodDescription.Type.ADDITION)
     public RecipeElvenTrade add(ItemStack output, IIngredient[] inputs) {
-        return add(new ItemStack[]{output}, inputs);
+        return recipeBuilder().input(inputs).output(output).register();
     }
```
brachy84, line 47: Array inputs are not good for groovy. Make it var args or collections/iterables

brachy84 (changes requested):
Awesome stuff!

brachy84 (approved):

PR #204, opened 2024-07-18
Add mixin to OreIngredient to access its name

juraj-hrivnak:
When using `crafting.streamRecipes()` and iterating through the recipes, it is impossible to get the name of an already existing oredict ingredient to be able to use it.

This PR adds an `OreIngredient.getOreDict()` method so you can access it in Groovy.
Example usage:

```groovy
crafting.streamRecipes().each { recipe ->
    String result = null

    recipe.ingredients.each { ingredient ->
        if (ingredient instanceof OreIngredient) {
            result = IngredientHelper.asGroovyCode(ingredient.oreDict, false)
        }
    }

    log.info(result)
}
```


Wizzerinus:
also useful for mod compats tbh. I've had an issue recently where I had to make a custom wrapper around a mod's recipe because OreIngredient doesn't retain the ore dict key.

with this code though, I think OreIngredient has multiple constructors, and only one of them properly keeps the oredict name..?

WaitingIdly:
your example usage indicates it is using `OreDictIngredient`, which is added by Groovyscript.

`OreIngredient` has a single constructor which does not save the oredict
image

juraj-hrivnak:
My bad, the example should use the `OreIngredient`. I edited it.

Yes, the `OreIngredient` constructor does not save the oredict, but it saves the oredict name, which is sufficient to convert it to the Groovy’s `OreDictIngredient`.

brachy84 (approved):

WaitingIdly (approved):

PR #205, opened 2024-07-23
Fix Edge Case Running Status of Sandbox

IntegerLimit:
This PR fixes edge cases where the running status of the GroovyScript sandbox is reported, falsely, as not running.

Reproduction Case:
```groovy
import gregtech.integration.groovy.GroovyScriptModule
import net.minecraft.item.ItemStack

println('Hello World!')

println(GroovyScriptModule.isCurrentlyRunning())

var test = { ItemStack stack ->
    println("Hello World (2)! " + stack.toString())
}

test(item('minecraft:apple'))

println(GroovyScriptModule.isCurrentlyRunning())
```

This script uses GT just to report the status of the sandbox, but it should work without gt dependency. This was reproduced in an instance with just GroovyScript and its deps, + GT and its deps.

Output of running the above script:
```
[21:59:45] [SERVER/INFO] [groovyscript]: ========== Reloading Groovy scripts ==========
[21:59:45] [SERVER/INFO] [placeholdername]: Running scripts in loader 'postInit'
[21:59:45] [SERVER/INFO] [placeholdername]:  - running main
Hello World!
true
Hello World (2)! 1xitem.apple@0
false
```

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/sandbox/GroovyScriptSandbox.java
+++ b/src/main/java/com/cleanroommc/groovyscript/sandbox/GroovyScriptSandbox.java
@@ -178,7 +178,7 @@ public class GroovyScriptSandbox extends GroovySandbox {
     @ApiStatus.Internal
     @Override
     public <T> T runClosure(Closure<T> closure, Object... args) {
-        startRunning();
+        if (!isRunning()) startRunning();
         T result = null;
         try {
             result = runClosureInternal(closure, args);
@@ -189,7 +189,7 @@ public class GroovyScriptSandbox extends GroovySandbox {
                 return new AtomicInteger();
             }).addAndGet(1);
         } finally {
-            stopRunning();
+            if (!isRunning()) stopRunning();
         }
         return result;
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/cleanroommc/groovyscript/sandbox/GroovyScriptSandbox.java
```diff
@@ -189,7 +189,7 @@ public <T> T runClosure(Closure<T> closure, Object... args) {
                 return new AtomicInteger();
             }).addAndGet(1);
         } finally {
-            stopRunning();
+            if (!isRunning()) stopRunning();
```
Wizzerinus, line 192: this permanently makes it running

Wizzerinus (commented):
I think that,

brachy84:
Wtf? How does *that* fix your issue?

src/main/java/com/cleanroommc/groovyscript/sandbox/GroovyScriptSandbox.java
IntegerLimit: Uhhh whoops copied my labs fix wrong...

IntegerLimit:
Fixing review comment now. I believe the fix still works with the proper fix, but will recheck.

brachy84:
Its closures... When they run running is always set to false

IntegerLimit:
> Its closures... When they stop running is always set to false

Breaks for closures... in scripts themselves, outside of event handlers, etc.

IntegerLimit:
Fix tested, works as intended.

IntegerLimit:
Essentially, here's why I think this issue occurs:
Closures running inside scripts still have that `runClosure` method called to run them.
When the closure finishes, `running` is set to false, therefore breaking the status.

Closures running in scripts are... pretty common.

brachy84 (approved):

PR #210, opened 2024-07-27
Resolve Symbolic Links in Groovy Folder in Debug Mode

IntegerLimit:
This PR properly resolves Symbolic Links when searching through the `groovy` folder to find scripts. This means that sub folders and individual files of the `groovy` folder can now be symlinked (`postInit` folder or folders inside `postInit`, etc.), rather than having the whole `groovy` folder required to be symlinked beforehand.

This results in potential improvements for developers.

This might introduce cases where recursive symlinks cause an infinite walk, which could be controlled with a limit on directory search, but given the rarity of such a case, and potential issues directory search limits could introduce, this has not been directly added in this PR.


IntegerLimit:
Change: Only followed in Debug Mode.

brachy84 (approved):

PR #220, opened 2024-08-12
Crl fixes and more

brachy84:
some cleanup, some fixes, some this, some that


.github/workflows/test_build.yaml
```diff
@@ -8,6 +8,7 @@ on:
   push:
     branches:
       - master
+      - crl-debug
```
WaitingIdly, line 11: should this get change get merged?

WaitingIdly (approved):

PR #221, opened 2024-08-13
Convert all Records to Classes

IntegerLimit:
This PR converts all usages of Records in GroovyScript to use Classes instead, preserving the same functions as before.

This is because these records were not compiled correctly, causing occasional GrS errors on some machines in some environments.


WaitingIdly (changes requested):
there are a number of files that still import `Desugar` - should optimize imports to ensure those are removed.

brachy84 (approved):

WaitingIdly (approved):

PR #231, opened 2024-08-28
fix singleton aspect, change texture, make example

### LINKED ISSUE #228: Cannot create basic aspect of Thaumcraft
If I don't fill in any sub-aspects, debugging will show an array out-of-bounds exception.

WaitingIdly:
changes in this PR:
- fixes aspect creation when there are no component aspects.
- adds an example for this situation.
- changes the texture and name so that it points to a valid texture and is not missingno.

resolves #228


src/main/java/com/cleanroommc/groovyscript/compat/mods/thaumcraft/aspect/Aspect.java
```diff
@@ -14,7 +14,10 @@
 @RegistryDescription
 public class Aspect extends VirtualizedRegistry<thaumcraft.api.aspects.Aspect> {
 
-    @RecipeBuilderDescription(example = @Example(".tag('humor').chatColor(14013676).component(aspect('cognitio')).component('perditio').image(resource('thaumcraft:textures/aspects/humor.png'))"))
+    @RecipeBuilderDescription(example = {
+            @Example(".tag('clay').chatColor(14013676).image(resource('placeholdername:textures/items/clay_2.png'))"),
+            @Example(".tag('snack').chatColor(14013676).component(aspect('cognitio')).component('clay').image(resource('placeholdername:textures/items/snack.png'))")
```
brachy84, line 19: should these color ints be hex ints?

WaitingIdly: good point, it would probably be more readable as hex codes - ill change those

PR #234, opened 2024-08-30
Boilerplate Registry for Lists

WaitingIdly:
changes in this PR:
- make `addBackup` and `addScripted` return a boolean in `VirtualizedRegistry` (always did in `AbstractReloadableStorage`).
- create a new class, `StandardListRegistry`, which extends `VirtualizedRegistry`, and contains a number of boilerplate methods for manipulating recipes in a list format. Note that while this refers to lists, anything that implements `Collection` will work.
- replaced 141 instances with the new `StandardListRegistry`, and adjusted code correspondingly.
- add some javadocs to the new class, along with `VirtualizedRegistry`, `ForgeRegistryWrapper`, and `NamedRegistry`.
- fix class name paths not recognizing underscores as a valid character in package/class names.
- fix a new line be added to examples if there are any methods, regardless of if any of the methods have any examples.
- make Better With Mods Heat compat have a type of `ENTRIES` instead of the default `RECIPES`.
- adjust the lang keys for Astral Sorcery's Fountain to reflect the new changes.
- added the ability for the documentation to handle generics via a `Map<generic name, real class>` map.
- add a pair of lang keys for the now-documented addition and removal methods for lists.
- side effect of these changes is that Better With Mods Heat, Mekanism Metallurgic Infuser, Mekanism Pressurized Reaction Chamber, Sawmill, LazyAE2 Fluix Aggregation, LazyAE2 Pulse Centrifuge, LazyAE2 Crystal Energization, LazyAE2 ME Circuit Etching now have methods they were previously missing.
- update examples with these changes.

this should both significantly reduce the number of lines in any relevant class and also standardize the interactions.


brachy84 (approved):

brachy84 (approved):
lgtm

PR #236, opened 2024-08-31
Make the Comp annotation easier to use

WaitingIdly:
changes in this PR:
- reworks how the `@Comp` annotation works with the goal of making it easier to write.
- deprecates the `type` and `value` elements of `@Comp`, and deprecates the `valid` and `requirement` elements of `@Property`.
- the old format works unchanged from its prior state, so this is not a breaking change.
- new format changes:
	- instead of an array of `@Comp` annotations, you now have a single `@Comp` annotation with multiple elements.
	- information is stored in different elements instead of a single `String` `value` element.
	- instead of a single `type` element, there is a `types` element.
	- if all of the required different elements have a non-default value, the `types` element can be omitted.
	- as a result of this, `types` is not required for any `@Comp` in our current codebase.
	- instead of using `Property#requirement`, `Comp#unique` is used.
	- `not` and `unique` elements are a `String`, the other elements are `int`.
	- this makes the format significantly less wordy - depending on the situation, it could be over 50% less stuff.
- joins via serial comma when more than 2 logical chunks are present.
- don't use `Integer.MAX_VALUE` in a `@Comp`, as its redundant in it being an `int` type.
- add more lang keys representing formerly `value` elements that are now migrated to be `unique` elements.
- make the `Comp.Type` enum declare the lang key in full.

some examples of what this looks like
```diff
-@Property(property = "input", valid = @Comp("1"))
+@Property(property = "input", comp = @Comp(eq = 1))
-@Property(property = "input", valid = {@Comp(value = "4", type = Comp.Type.GTE), @Comp(value = "5", type = Comp.Type.LTE)})
+@Property(property = "input", comp = @Comp(gte = 4, lte = 5))
-@Property(valid = {@Comp(type = Comp.Type.GTE, value = "0"), @Comp(type = Comp.Type.LTE, value = "1")})
+@Property(comp = @Comp(gte = 0, lte = 1))
-@Property(value = "groovyscript.wiki.craftingrecipe.ingredientMatrix.value", requirement = "groovyscript.wiki.craftingrecipe.matrix.required", valid = {@Comp(value = "1", type = Comp.Type.GTE), @Comp(value = "9", type = Comp.Type.LTE)}, priority = 200, hierarchy = 20)
+@Property(value = "groovyscript.wiki.craftingrecipe.ingredientMatrix.value", comp = @Comp(gte = 1, lte = 9, unique = "groovyscript.wiki.craftingrecipe.matrix.required"), priority = 200, hierarchy = 20)
```


brachy84 (approved):
probably fine

PR #240, opened 2024-09-04
Artisan Worktables compat, wiki builder improvements

Wizzerinus:
This PR adds the following:
- Compat for Artisan Worktables mod
- `raw` toggle on `@Example`, which will currently not add `.recipeBuilder()` to the method description on the wiki
- `virtual` toggle on `@Property`, which allows adding `@Property` not represented by a field in the target class
- Cleans up the mess after Immersive Tech PR in gradle config

Should be ready for a review now


src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+            return "Error adding Artisan Worktable recipe";
+        }
+
+        @Override
+        public void validate(GroovyLog.Msg msg) {
+            if (name == null || name.isEmpty()) {
+                name = RecipeName.generate("groovyscript_");
+            }
+
+            validateItems(msg, 0, 9, 0, 0);
```
WaitingIdly, line 443: i think there needs to be at least one input, right?

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+        @RecipeBuilderMethodDescription
+        public ArtisanRecipeBuilder name(ResourceLocation name) {
+            return this.name(name.toString());
+        }
+
+        @RecipeBuilderMethodDescription
+        public ArtisanRecipeBuilder mirrored(boolean mirrored) {
+            if (!shaped) {
+                issues.add("Shapeless recipes cannot be mirrored");
+            }
```
WaitingIdly, line 201: this could be handled in `validate`

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+            return this;
+        }
+
+        @RecipeBuilderMethodDescription
+        public ArtisanRecipeBuilder name(ResourceLocation name) {
+            return this.name(name.toString());
+        }
+
+        @RecipeBuilderMethodDescription
+        public ArtisanRecipeBuilder mirrored(boolean mirrored) {
```
WaitingIdly, line 198: add a method that inverts the mirrored status
```java
        @RecipeBuilderMethodDescription
        public ArtisanRecipeBuilder mirrored() {
            this.mirrored = !mirrored;
            return this;
        }

File: src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java [hunk 4]
```diff
@@ -0,0 +1,458 @@
...
+        @Property
+        private int level = 0;
+
+        @Property(defaultValue = "true")
+        private boolean consumeExperience = true;
+
+        @Property(defaultValue = "empty", valid = @Comp(type = Comp.Type.LTE, value = "3"))
+        private final List<ExtraOutputChancePair> extraOutputs = new ArrayList<>();
+
+        @Property(defaultValue = "the lowest tier that can hold all inputs of this recipe")
```
WaitingIdly, line 161: i think this is better served as part of the description

src/main/java/com/cleanroommc/groovyscript/documentation/Builder.java
```diff
@@ -321,7 +335,9 @@ private String createBuilder(Example example, boolean canBeCommented) {
 
         if (!example.def().isEmpty()) out.append("def ").append(example.def()).append(" = ");
 
-        out.append(reference).append(".").append(builderMethod.getName()).append("()");
+        out.append(reference);
+
+        if (!example.raw()) out.append(".").append(builderMethod.getName()).append("()");
```
WaitingIdly, line 340: if this is still needed (im not seeing why it needs to be in the constructor, think thats a relic of the passthrough logic):

would prefer instead having something like this: 
```java
out.append(reference).append(".").append(builderMethod.getName());
out.append("(");
if (!example.params().isEmpty()) out.append(example.params());
out.append(")");
```
where `params` is a `String` with a default of `""`.
as an example, this would do this:
```diff
-@Example(raw = true, value = ".shapelessBuilder('basic').gridInput(item('minecraft:coal'), item('minecraft:iron_ingot')).output(item('minecraft:clay')).maximumTier(1).minimumTier(1)")
+@Example(params = "'basic'", value = ".gridInput(item('minecraft:coal'), item('minecraft:iron_ingot')).output(item('minecraft:clay')).maximumTier(1).minimumTier(1)")
```
another option is to make `value` be an array of strings, and then for each parameter of the method grab the next entry in value. probably overengineering that one.

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+            return this;
+        }
+
+        @RecipeBuilderMethodDescription
+        public ArtisanRecipeBuilder output(ItemStack stack) {
+            return output(stack, 1);
+        }
+
+        @RecipeBuilderMethodDescription
+        public ArtisanRecipeBuilder output(ItemStack stack, int weight) {
```
WaitingIdly, line 256: can there be a method to add a `OutputWeightPair` directly?

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+            return this;
+        }
+
+        @RecipeBuilderMethodDescription(field = "extraOutputs")
+        public ArtisanRecipeBuilder extraOutput(ItemStack stack) {
+            return extraOutput(stack, 1.0f);
+        }
+
+        @RecipeBuilderMethodDescription(field = "extraOutputs")
+        public ArtisanRecipeBuilder extraOutput(ItemStack stack, float chance) {
```
WaitingIdly, line 267: can there be a method to add a `ExtraOutputChancePair` directly?

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+
+        @RecipeBuilderMethodDescription
+        public ArtisanRecipeBuilder maximumTier(int maximumTier) {
+            this.maximumTier = maximumTier;
+            return this;
+        }
+
+        @RecipeBuilderMethodDescription
+        public ArtisanRecipeBuilder hidden() {
+            return hidden(true);
```
WaitingIdly, line 286: should invert rather than set to be true

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+            }
+            GroovyLog.Msg message =
+                    GroovyLog.msg("Error adding Artisan Worktables recipe").error()
+                    .add(tier == null, "Unable to calculate the table tier");
+            if (tier != null) {
+                int tierNeeded = tier.getId();
+                message.add(tierNeeded > maximumTier, "The recipe requires at least tier {}, but maximumTier was set to {}", tierNeeded, maximumTier);
+                minimumTier = Math.max(minimumTier, tierNeeded);
+            }
+            if (message.postIfNotEmpty()) return null;
```
WaitingIdly, line 401: this should get moved to validate

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+
+            // The mod expects three Nonnull ExtraOutputChancePairs... yuck
+            ExtraOutputChancePair[] extraOutputs = new ExtraOutputChancePair[3];
+            Arrays.fill(extraOutputs, new ExtraOutputChancePair(ArtisanItemStack.EMPTY, 0));
+            for (int i = 0; i < this.extraOutputs.size(); i++) extraOutputs[i] = this.extraOutputs.get(i);
+
+            IArtisanRecipe recipe = new ArtisanRecipe(
+                    this.makeArtisanName(),
```
WaitingIdly, line 409: could do `validateName() and do `name.toString()` here - it looks like the name is just a `ResourceLocation#toString`

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+                    requirements,
+                    output,
+                    tools.toArray(new ToolEntry[0]),
+                    data.ingredients.stream().map(r -> ArtisanIngredient.from(r.toMcIngredient())).collect(Collectors.toList()),
+                    input.stream().map(r -> ArtisanIngredient.from(r.toMcIngredient())).collect(Collectors.toList()),
+                    consumeSecondaryIngredients,
+                    fluidInput.isEmpty() ? null : fluidInput.get(0),
```
WaitingIdly, line 416: `fluidInput.getOrDefault(0, null)` works, right?

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+        }
+
+        public ArtisanRecipeBuilder(String name, boolean shaped) {
+            this.registryName = name;
+            this.shaped = shaped;
+            if (shaped) {
+                this.gridBuilder = new ShapedGridBuilder();
+            } else {
+                this.gridBuilder = new ShapelessGridBuilder();
+            }
```
WaitingIdly, line 183: theres a lot of "oh cant do this if shaped" or "not shapeless!" - i really think you should split it up. it will be a bit of repetition, but not that much - and an interface can enforce the methods

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+        }
+
+        @RecipeBuilderMethodDescription(field = "tools")
+        public ArtisanRecipeBuilder tool(IIngredient tool, int durability) {
+            tools.add(new ToolEntry(ArtisanIngredient.from(tool.toMcIngredient()), durability));
+            return this;
+        }
+
+        @RecipeBuilderMethodDescription
+        public ArtisanRecipeBuilder consumeSecondaryIngredients(boolean consume) {
```
WaitingIdly, line 227: can this get an inversion method?

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+        }
+
+        @RecipeBuilderMethodDescription
+        public ArtisanRecipeBuilder level(int level) {
+            this.level = level;
+            return this;
+        }
+
+        @RecipeBuilderMethodDescription
+        public ArtisanRecipeBuilder consumeExperience(boolean consume) {
```
WaitingIdly, line 245: can this get an inversion method?

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+        private boolean hidden = false;
+
+        private final AbstractCraftingRecipeBuilder<ArtisanRecipeData> gridBuilder;
+
+        private String makeArtisanName() {
+            return this.registryName + ':' + this.name;
+        }
+
+        public ArtisanRecipeBuilder(String name, boolean shaped) {
+            this.registryName = name;
```
WaitingIdly, line 177: i dont see why this has to be set via the constructor - its only referenced in `validate()` and `register()`

WaitingIdly (changes requested):
i believe there was a change in development that means the constructor no longer needs the table name passed in. if it is no longer needed, the related change to the documentation annotation should be removed.

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
Wizzerinus: Like a method with no arguments?

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
Wizzerinus: it's 12 methods that are the same, with a bunch of annotations. copying 12 methods violates every single principle of programming :D

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+                this.requirements.put(location, matchRequirementBuilder.create());
+            } else {
+                issues.add(String.format("Mod %s required for %s was not loaded", modIdRequired, matchRequirementBuilder));
+            }
+
+            return this;
+        }
+
+        @RecipeBuilderMethodDescription(field = "tools")
+        public ArtisanRecipeBuilder tool(IIngredient tool, int durability) {
```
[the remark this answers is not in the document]
Wizzerinus: not sure the usecases for that but I'll allow i

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
Wizzerinus: the inputs is for *secondary* inputs, not for the grid inputs. Each composited class validates grid inputs itself.

src/main/java/com/cleanroommc/groovyscript/documentation/Builder.java
Wizzerinus: can do the first one, yeah

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
Wizzerinus: its not a resource location, it has to have the table ID as the resource location's namespace, like `mage:recipe_id`

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
Wizzerinus: this makes sense, yeah

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+        return new ArtisanRecipeBuilder(name, false);
+    }
+
+    static class ArtisanRecipeData {
+        List<IIngredient> ingredients;
+        int width;
+        int height;
+    }
+
+    static class ShapedGridBuilder extends AbstractCraftingRecipeBuilder.AbstractShaped<ArtisanRecipeData> {
```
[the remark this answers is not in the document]
Wizzerinus: I mean it doesn't have to be public...

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
```diff
@@ -0,0 +1,458 @@
...
+
+@RegistryDescription(admonition = {
+        @Admonition(type = Admonition.Type.INFO, value = "groovyscript.wiki.artisanworktables.tabletypes"),
+        @Admonition(type = Admonition.Type.DANGER, value = "groovyscript.wiki.artisanworktables.ctcrash")
+})
+public class Tables extends VirtualizedRegistry<IArtisanRecipe> {
+
+    @Override
+    public void onReload() {
+        removeScripted().forEach(r -> {
```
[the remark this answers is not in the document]
Wizzerinus: I mean, I think adding any `removeRecipe` is too complicated because the recipes have multiple inputs of three different types, multiple outputs of two different types and so on. So the backup will always be empty

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
Wizzerinus: Precedent for the current implementation btw:

image

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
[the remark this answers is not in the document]
Wizzerinus: made all of them private instead of having the default visibility

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
Wizzerinus: the issue I have with that is it will silently allow setting `mirrored(false)`

src/main/java/com/cleanroommc/groovyscript/documentation/Builder.java
Wizzerinus: Thinking about it I no longer need the feature. I could just remove the raw, I guess, but I feel like it's fine to keep for housekeeping.

src/main/java/com/cleanroommc/groovyscript/compat/mods/artisanworktables/Tables.java
Wizzerinus: no because it's not a list

Wizzerinus:
done

src/main/java/com/cleanroommc/groovyscript/api/documentation/annotations/RecipeBuilderDescription.java
```diff
@@ -51,4 +51,11 @@
      */
     int priority() default 1000;
 
+    /**
+     * The properties that should not be documented in this recipe builder, even though they exist in the class.
+     *
+     * @return the array of skipped properties
+     */
+    String[] skippedProperties() default {};
```
WaitingIdly, line 59: this already exists via `@Property(property = "thing", needsOverride = true)`

Wizzerinus: That's different. This is specifically attached to the Recipe Builder instead of its class to handle the case with having the same class. Although, thinking about it, the cleanest solution might be to have WorktableRecipeBuilderBase that gets inherited twice to obtain Shaped and Shapeless versions? That might be better than constantly checking for if(shaped). Let me know if I should change to that.

WaitingIdly: you can put property annotations inside the builder documentation annotation

Wizzerinus:
ok I did it, sorry for the delay - I was going through Things

Wizzerinus:
Could I get a re-review on this 🥺

examples/postInit/artisanworktables.groovy
```diff
@@ -0,0 +1,41 @@
...
+    .output(item('minecraft:clay'), item('minecraft:nether_star'))
+    .register()
+
+mods.artisanworktables.tables.shapelessBuilder()
+    .type('basic')
+    .gridInput(item('minecraft:coal'), item('minecraft:iron_ingot'))
+    .output(item('minecraft:clay'))
+    .maximumTier(1)
+    .minimumTier(1)
+    .register()
```
WaitingIdly, line 39: i would suggest copying some of the existing grid-based recipe examples and slightly modifying them - there should definitely be more than 3 examples for such a complex builder.

WaitingIdly (changes requested):
the builder classes need to be split up. its horrible to read and navigate. documenting fields that dont exist via `virtual` seems like a good thing to happen, but it doesnt work properly and there is not a good way to make it work properly.
to be clear: the current documentation does not generate, and the only way to get it to generate properly would require at least two more elements added to the `@Property` annotation (and require very specific class declaration stuff) - exclusively to do this.

please scrap the DRY and just WET.

Wizzerinus:
I really don't like the current structure of this, don't see an easy way to improve the structure, and don't have time to come up with a better way, so I'm going to close this now. Sorry! </3

PR #242, opened 2024-10-01
cant download, wont use

WaitingIdly:
changes in this PR:
- since Advanced Rocketry 2 restricts downloads for third parties (a recent change), we cant have it as a dependency. gootbye


brachy84:
gootbye

PR #254, opened 2024-10-05
add Future MC compat

WaitingIdly:
changes in this PR:
- adds Future MC `stonecutter`, `smithing`, `campfire`, `blastFurnace`, and `smoking` compat. adds examples and i18n for those.


brachy84 (approved):
The recipes are all very simple. Maybe add simple `add(...)` methods?

PR #258, opened 2024-10-11
add Ice And Fire compat

WaitingIdly:
changes in this PR:
- adds compat with Ice And Fire `FireForge` and `IceForge`. examples and i18n for those.
- adds compat with Ice And Fire: RotN Edition for the `LightningForge`, with examples that default to being commented out and i18n explaining the situation
- moves the complex industrialcraft dependencies out of the map.
- makes example comments always be line comments instead of sometimes being block comments


src/main/java/com/cleanroommc/groovyscript/compat/mods/iceandfire/IceAndFire.java
```diff
@@ -0,0 +1,25 @@
...
+    public final IceForge iceForge = new IceForge();
+    public final LightningForge lightningForge;
+
+    public IceAndFire() {
+        lightningForge = isRotN() ? new LightningForge() : null;
+    }
+
+    public static boolean isRotN() {
+        for (var container : Loader.instance().getActiveModList()) {
+            if ("iceandfire".equals(container.getModId())) {
```
brachy84, line 18: you can get mod container from a map. Its something like `Loader.instance().getIndexedModList()`

brachy84 (changes requested):

brachy84 (approved):

PR #269, opened 2024-11-09
Commands helper

brachy84:
- adds global `commands` variable which has helper methods to register client and server commands
- if commands are registered during `postInit` they can be reloaded
- see test in `custom/vanilla.groovy`

### DIFF
--- a/examples/postInit/custom/vanilla.groovy
+++ b/examples/postInit/custom/vanilla.groovy
@@ -256,3 +256,7 @@ item('minecraft:golden_apple').setRarity(textformat('-1'))
 eventManager.listen(EnderTeleportEvent) { event ->
     event.setAttackDamage 19.5f
 }
+
+commands.registerCommand('groovy_test') { server, sender, args ->
+    sender.sendMessage('Hello from GroovyScript')
+}
--- a/src/main/java/com/cleanroommc/groovyscript/compat/vanilla/Commands.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/vanilla/Commands.java
@@ -1,0 +1,99 @@
+package com.cleanroommc.groovyscript.compat.vanilla;
+
+import com.cleanroommc.groovyscript.GroovyScript;
+import com.cleanroommc.groovyscript.api.GroovyBlacklist;
+import com.cleanroommc.groovyscript.api.IScriptReloadable;
+import com.cleanroommc.groovyscript.command.SimpleCommand;
+import com.cleanroommc.groovyscript.core.mixin.CommandHandlerAccessor;
+import com.cleanroommc.groovyscript.registry.AbstractReloadableStorage;
+import com.cleanroommc.groovyscript.registry.NamedRegistry;
+import net.minecraft.command.CommandHandler;
+import net.minecraft.command.ICommand;
+import net.minecraft.server.MinecraftServer;
+import net.minecraftforge.client.ClientCommandHandler;
+import net.minecraftforge.fml.common.FMLCommonHandler;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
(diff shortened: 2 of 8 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/vanilla/Commands.java
```diff
@@ -0,0 +1,99 @@
...
+        } else {
+            this.serverCommands.add(command);
+        }
+        if (this.serverStarted) {
+            forServer(commandHandler -> commandHandler.registerCommand(command));
+        }
+    }
+
+    public void registerClientCommand(ICommand command) {
+        if (FMLCommonHandler.instance().getSide().isServer()) return;
```
WaitingIdly, line 41: should this print a warning?

src/main/java/com/cleanroommc/groovyscript/compat/vanilla/Commands.java
```diff
@@ -0,0 +1,99 @@
...
+
+public class Commands extends NamedRegistry implements IScriptReloadable {
+
+    private final List<ICommand> serverCommands = new ArrayList<>();
+    private final AbstractReloadableStorage<ICommand> serverReloadableCommands = new AbstractReloadableStorage<>();
+    private final AbstractReloadableStorage<ICommand> clientReloadableCommands = new AbstractReloadableStorage<>();
+    private boolean serverStarted = false;
+
+    public void registerCommand(ICommand command) {
+        if (GroovyScript.getSandbox().isRunning() && GroovyScript.getSandbox().getCurrentLoader().isReloadable()) {
```
WaitingIdly, line 30: i feel like this code being required for `Commands` to work in preInit (if im reading what the reason is correctly) might indicate a deeper issue.

src/main/java/com/cleanroommc/groovyscript/compat/vanilla/Commands.java
```diff
@@ -0,0 +1,99 @@
...
+import net.minecraftforge.client.ClientCommandHandler;
+import net.minecraftforge.fml.common.FMLCommonHandler;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Consumer;
+
+public class Commands extends NamedRegistry implements IScriptReloadable {
```
WaitingIdly, line 22: i think this should be `Command` in the singular

WaitingIdly (changes requested):
it might be for documentation only (yippee), but doing something like `commands.registerCommand('help') { server, sender, args -> sender.sendMessage('You cannot be helped') }`
lets you override commands - and, depending on when things are loaded, might make your custom commands be overridden by preexisting ones.

src/main/java/com/cleanroommc/groovyscript/compat/vanilla/Commands.java
brachy84: No. This way you can just register a command  without having to make sure it doent run on server

src/main/java/com/cleanroommc/groovyscript/compat/vanilla/Commands.java
brachy84: ? I just dont want commands to reload when not registered durin postInit. I dont see any issue

examples/postInit/custom/vanilla.groovy
```diff
@@ -256,3 +256,7 @@ item('minecraft:golden_apple').setRarity(textformat('-1'))
 eventManager.listen(EnderTeleportEvent) { event ->
     event.setAttackDamage 19.5f
 }
+
+commands.registerCommand('groovy_test') { server, sender, args ->
```
zznty, line 260: should be `command.registerCommand` after rename

PR #274, opened 2024-12-03
Fix EnderIO Inventory Panel Crafting Crash

WaitingIdly:
changes in this PR:
- check that the inventory is an instanceof `InventoryCraftResult` before casting to it.

this resolves https://github.com/FTBTeam/FTB-Modpack-Issues/issues/6310


brachy84:
temporary fix

PR #279, opened 2024-12-05
add Default GameRules

WaitingIdly:
changes in this PR:
- add the ability to set the default GameRules. can be done one-by-one or en-mass. 
- when a gamerule that doesnt exist tries to be added, log a warning unless the user has run `gameRule.setWarnNewGameRule(false)`.
- add a command `/gs applyDefaultGameRules` to apply these in-game.
- add examples for this
- the default gamerules are applied when `WorldEvent.CreateSpawnPosition` is fired - as far as i can tell, it is only fired once, when the world is created. other places that do similar things listen to `FMLServerStartingEvent` (applies on every load), `WorldEvent.Load` (applies on every load), and a particularly daring one did `EntityJoinWorldEvent`. 
	- i tested `WorldEvent.CreateSpawnPosition` on both client and server, and it only runs if the `level.dat` does not exist (something which tripped me up on the server for a bit due to always terminating the run instead of stopping the server).
- fix the server using the `run/examples` folder instead of `examples` when `groovyscript.use_examples_folder` is set to true. bug only affects us.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/compat/vanilla/GameRule.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/vanilla/GameRule.java
@@ -5,15 +5,15 @@ import com.cleanroommc.groovyscript.api.GroovyLog;
 import com.cleanroommc.groovyscript.api.IScriptReloadable;
 import com.cleanroommc.groovyscript.api.documentation.annotations.MethodDescription;
 import com.cleanroommc.groovyscript.registry.NamedRegistry;
+import it.unimi.dsi.fastutil.objects.Object2ObjectOpenHashMap;
 import net.minecraft.world.GameRules;
 
-import java.util.HashMap;
 import java.util.Map;
 
 public class GameRule extends NamedRegistry implements IScriptReloadable {
 
     private static final String LOG_MESSAGE = "Could not find an already existing rule with the name {}. This may be intentional! If it is, you can disable this via `gameRule.setWarnNewGameRule(false)`";
-    private final Map<String, String> defaultGameRules = new HashMap<>();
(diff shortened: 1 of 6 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/vanilla/GameRule.java
```diff
@@ -0,0 +1,54 @@
...
+import com.cleanroommc.groovyscript.registry.NamedRegistry;
+import net.minecraft.world.GameRules;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class GameRule extends NamedRegistry implements IScriptReloadable {
+
+    private static final String LOG_MESSAGE = "Could not find an already existing rule with the name {}. This may be intentional! If it is, you can disable this via `gameRule.setWarnNewGameRule(false)`";
+    private final Map<String, String> defaultGameRules = new HashMap<>();
```
brachy84, line 16: fastutil map?

brachy84 (changes requested):
epic

src/main/java/com/cleanroommc/groovyscript/compat/vanilla/GameRule.java
WaitingIdly: uh, its object to object, i was under the understanding that either `Object2ObjectOpenHashMap` or `HashMap` are fine, and theres little difference. is this incorrect?

brachy84: fastutil is always the better choice for hash maps

brachy84 (approved):

PR #280, opened 2024-12-08
Mapper backend refactor and texture deco improvements

brachy84:
- add an `AbstractObjectMapper<T>` class
- mappers can now be created using the builder or by extending the new class
- texture decos should now be thread safe (i dont really now what im doing there)
- some code structure improvements for texture decos
- some improvements for texture decos to avoid creating them multiple times
- this branch was successfully tested for backwards compat with gt


src/main/java/com/cleanroommc/groovyscript/server/features/textureDecoration/TextureDecorationProvider.java
```diff
@@ -33,23 +29,22 @@
...
 import java.util.concurrent.CompletableFuture;
-import java.util.function.Supplier;
 
 public class TextureDecorationProvider extends DocProvider {
 
-    private static final int ICON_W = 16, ICON_H = 16;
-    private static final int ICON_X = 0, ICON_Y = 0;
-    private static final Map<String, TextureDecoration> textures = new Object2ObjectOpenHashMap<>();
+    public static final int ICON_W = 16, ICON_H = 16;
+    public static final int ICON_X = 0, ICON_Y = 0;
```
WaitingIdly, line 36: can these be split into separate lines?

src/main/java/com/cleanroommc/groovyscript/mapper/TooltipEmbedding.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.apache.commons.codec.digest.DigestUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Pattern;
+
+public class TooltipEmbedding<T> {
+
+    private static final Pattern embeddingPattern = Pattern.compile("(?>\\$\\{(?<mapper>\\w+)\\(['\\\"](?<key>[\\w\\s:-]+)['\\\"]\\)})");
```
WaitingIdly, line 11: intellij tells me that `\\\"` can be shortened to `\"`

src/main/java/com/cleanroommc/groovyscript/server/features/textureDecoration/TextureDecoration.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class TextureDecoration<T> {
+
+    private static <T> T getObjectWithConstArgs(AbstractObjectMapper<T> mapper, ArgumentListExpression args) {
+        var additionalArgs = new Object[args.getExpressions().size() - 1];
+        for (int i = 0; i < additionalArgs.length; i++) {
+            if (args.getExpressions().get(i + 1) instanceof ConstantExpression argExpression)
+                additionalArgs[i] = argExpression.getValue();
```
WaitingIdly, line 24: i think this should get wrapped in curly braces

src/main/java/com/cleanroommc/groovyscript/server/features/textureDecoration/TextureDecoration.java
```diff
@@ -0,0 +1,108 @@
...
+        return new TextureDecoration<>(embedding.getTextureName(), embedding.getMapper(), embedding.getContext(), TextureDecorationProvider.getURIForDecoration(embedding.getTextureName()));
+    }
+
+    private TextureDecoration(String name, AbstractObjectMapper<T> mapper, T bindable, String uri) {
+        this.name = name;
+        this.mapper = mapper;
+        this.bindable = bindable;
+        this.uri = uri;
+        File file = getFile();
+        this.fileExists = !file.getParentFile().mkdirs() && file.exists();
```
WaitingIdly, line 52: why is `!file.getParentFile().mkdirs()` needed?

src/main/java/com/cleanroommc/groovyscript/server/features/textureDecoration/TextureDecoration.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public void bindTexture() {
+        this.mapper.bindTexture(this.bindable);
+    }
+
+    public List<String> getTooltip() {
+        List<String> tooltip = this.mapper.getTooltip(this.bindable);
+        return tooltip.isEmpty() ? tooltip : new ArrayList<>(tooltip);
```
WaitingIdly, line 61: whats the difference between these? why `new ArrayList<>(tooltip)` if `tooltip` is empty?
```diff
-tooltip.isEmpty() ? tooltip : new ArrayList<>(tooltip)
+tooltip.isEmpty() ? tooltip : new ArrayList<>()
```

src/main/java/com/cleanroommc/groovyscript/server/features/textureDecoration/TextureDecoration.java
```diff
@@ -0,0 +1,108 @@
...
+        saveImage(getFile(), buffer);
+        buffer.rewind();
+        synchronized (this) {
+            this.fileExists = true;
+            this.queued = false;
+        }
+    }
+
+    private static void saveImage(File file, ByteBuffer buffer) {
+        final int w = TextureDecorationProvider.ICON_W, h = TextureDecorationProvider.ICON_H;
```
WaitingIdly, line 89: split into separate declarations

src/main/java/com/cleanroommc/groovyscript/mapper/AbstractObjectMapper.java
```diff
@@ -0,0 +1,165 @@
...
+    public void bindTexture(T t) {}
+
+    @NotNull
+    public List<String> getTooltip(T t) {
+        return Collections.emptyList();
+    }
+
+    public boolean hasTextureBinder() {
+        if (this.hasTextureBinder == null) {
+            for (Method method : getClass().getDeclaredMethods()) {
```
WaitingIdly, line 99: why does this need reflection?

src/main/java/com/cleanroommc/groovyscript/mapper/BlockStateMapper.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public @NotNull Result<IBlockState> parse(String mainArg, Object[] args) {
+        return ObjectMappers.parseBlockState(mainArg, args);
+    }
+
+    @Override
+    public void provideCompletion(int index, Completions items) {
+        if (index == 0) items.addAllOfRegistry(ForgeRegistries.BLOCKS);
```
WaitingIdly, line 35: is it possible to have completion for properties of blockstates?

src/main/java/com/cleanroommc/groovyscript/mapper/ItemStackMapper.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public @NotNull Result<ItemStack> parse(String mainArg, Object[] args) {
+        return ObjectMappers.parseItemStack(mainArg, args);
+    }
+
+    @Override
+    public void provideCompletion(int index, Completions items) {
+        if (index == 0) items.addAllOfRegistry(ForgeRegistries.ITEMS);
```
WaitingIdly, line 38: is it possible for the valid metadata of items to have autocompletion? pretty sure its a "no"

src/main/java/com/cleanroommc/groovyscript/mapper/BlockStateMapper.java
```diff
@@ -0,0 +1,49 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraftforge.fml.common.registry.ForgeRegistries;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Collections;
+import java.util.List;
+
+public class BlockStateMapper extends AbstractObjectMapper<IBlockState> {
+
+    public static final BlockStateMapper INSTANCE = new BlockStateMapper("blockstate", null);
```
WaitingIdly, line 17: since this is the only place you are going to init `BlockStateMapper`, why pass `null` here and not drop and argument and call `super(name, null, IBlockState.class)`?

src/main/java/com/cleanroommc/groovyscript/mapper/BlockStateMapper.java
```diff
@@ -0,0 +1,49 @@
...
+        super(name, mod, IBlockState.class);
+    }
+
+    @Override
+    public Result<IBlockState> getDefaultValue() {
+        return Result.some(Blocks.AIR.getDefaultState());
+    }
+
+    @Override
+    public @NotNull Result<IBlockState> parse(String mainArg, Object[] args) {
```
WaitingIdly, line 29: this and a few other places have switched from `parse(String mainArg, Object... args)` to `parse(String mainArg, Object[] args)` - while this makes no difference to groovy (or to overrides, apparently, which is interesting), is this intentional?
should we continue using varargs for things exposed to groovy, even if it makes no difference in functionality (although it does display differently on the wiki)?

WaitingIdly (changes requested):
this was an issue prior to this PR, but what should be done about animated textures, if anything?

src/main/java/com/cleanroommc/groovyscript/server/features/textureDecoration/TextureDecoration.java
brachy84: No. Return it if empty and copy if not.

WaitingIdly: right sorry misread, had that flipped in my head as "if its not empty return it, otherwise return an empty list".

src/main/java/com/cleanroommc/groovyscript/server/features/textureDecoration/TextureDecoration.java
zznty: at the time i was writing this, nothing was creating cache folder before so i check it here and set exists to false if directory did not exist prior to this call

src/main/java/com/cleanroommc/groovyscript/mapper/BlockStateMapper.java
brachy84: I thought about that, but we would need information about what the first param currently is

src/main/java/com/cleanroommc/groovyscript/mapper/ItemStackMapper.java
brachy84: It is possible, but id argue that its not worth it

src/main/java/com/cleanroommc/groovyscript/mapper/BlockStateMapper.java
brachy84: I was doing that for item so that mods can extend that class for a template (for example gregtech metaitems). I guess its not very useful for blockstate

src/main/java/com/cleanroommc/groovyscript/mapper/ItemStackMapper.java
brachy84: It was not possible. Not it might be. But idk how to trigger completion for ints

PR #281, opened 2024-12-08
Fix item transforms

brachy84:


src/main/java/com/cleanroommc/groovyscript/helper/ingredient/IngredientBase.java
```diff
@@ -1,49 +1,80 @@
...
+            return self;
+        });
+    }
+
     public IngredientBase reuse() {
-        return transform(IngredientHelper.REUSE);
+        return transform(self -> self);
+    }
+
+    public IngredientBase noreturn() {
```
WaitingIdly, line 53: should this be `noReturn`? if so, should probably keep `noreturn` and either have it be a duplicate or deprecate it.

src/main/java/com/cleanroommc/groovyscript/compat/vanilla/ItemStackMixinExpansion.java
```diff
@@ -220,6 +223,16 @@ default ItemStack withDamage(int meta) {
         return withMeta(meta);
     }
 
+    default ItemStack destroy() {
```
WaitingIdly, line 226: should this check if the item is already empty? the other place i see this being done does so.

src/main/java/com/cleanroommc/groovyscript/helper/ingredient/IngredientHelper.java
```diff
@@ -25,6 +27,10 @@ public class IngredientHelper {
     public static final Closure<Object> NO_RETURN = new LambdaClosure<>(args -> ItemStack.EMPTY);
     public static final Closure<Object> MATCH_NBT = new LambdaClosure<>(args -> ItemStack.EMPTY);
 
+    public static ItemStack damageItem(ItemStack stack, int damage) {
+        return ItemStackMixinExpansion.of(stack).withMeta(Math.min(32767, Items.DIAMOND.getDamage(stack) + damage));
```
WaitingIdly, line 31: could this method get some javadocs to explain this, particularly the `Items.DIAMOND` part? might be worth doing `Short.MAX_VALUE` instead of `32767`.

WaitingIdly (changes requested):

src/main/java/com/cleanroommc/groovyscript/compat/vanilla/ItemStackMixinExpansion.java
```diff
@@ -130,6 +129,10 @@ default ItemStack applyTransform(ItemStack matchedInput) {
         if (grs$getTransformer() != null) {
             ItemStack result = grs$getTransformer().transform(matchedInput);
             if (result == null) return ItemStack.EMPTY;
```
WaitingIdly, line 131: should this be 
```suggestion
            if (result == null || result.isEmpty()) return ItemStack.EMPTY;
```

WaitingIdly (approved):

PR #283, opened 2024-12-22
added lightningcraft compat

Wizzerinus:
literally no one asked for this but here we go

oh wait Lord_Chaos did

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/compat/mods/ModSupport.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/mods/ModSupport.java
@@ -42,6 +42,7 @@ import com.cleanroommc.groovyscript.compat.mods.inspirations.Inspirations;
 import com.cleanroommc.groovyscript.compat.mods.integrateddynamics.IntegratedDynamics;
 import com.cleanroommc.groovyscript.compat.mods.jei.JustEnoughItems;
 import com.cleanroommc.groovyscript.compat.mods.lazyae2.LazyAE2;
+import com.cleanroommc.groovyscript.compat.mods.lightningcraft.Lightningcraft;
 import com.cleanroommc.groovyscript.compat.mods.magneticraft.Magneticraft;
 import com.cleanroommc.groovyscript.compat.mods.mekanism.Mekanism;
 import com.cleanroommc.groovyscript.compat.mods.mysticalagriculture.MysticalAgriculture;
@@ -124,6 +125,7 @@ public class ModSupport {
     public static final GroovyContainer<Mekanism> MEKANISM = new InternalModContainer<>("mekanism", "Mekanism", Mekanism::new);
     public static final GroovyContainer<MysticalAgriculture> MYSTICAL_AGRICULTURE = new InternalModContainer<>("mysticalagriculture", "Mystical Agriculture", MysticalAgriculture::new);
     public static final GroovyContainer<LazyAE2> LAZYAE2 = new InternalModContainer<>("threng", "LazyAE2", LazyAE2::new, "lazyae2");
+    public static final GroovyContainer<Lightningcraft> LIGHTNINGCRAFT = new InternalModContainer<>("lightningcraft", "Lightningcraft", Lightningcraft::new);
     public static final GroovyContainer<NaturesAura> NATURES_AURA = new InternalModContainer<>("naturesaura", "Nature's Aura", NaturesAura::new);
     public static final GroovyContainer<PneumaticCraft> PNEUMATIC_CRAFT = new InternalModContainer<>("pneumaticcraft", "PneumaticCraft: Repressurized", PneumaticCraft::new);
     public static final GroovyContainer<PrimalTech> PRIMAL_TECH = new InternalModContainer<>("primal_tech", "Primal Tech", PrimalTech::new, "primaltech");
--- a/src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
@@ -1,0 +1,106 @@
+package com.cleanroommc.groovyscript.compat.mods.lightningcraft;
+
+import com.cleanroommc.groovyscript.api.GroovyLog;
+import com.cleanroommc.groovyscript.api.IIngredient;
+import com.cleanroommc.groovyscript.api.documentation.annotations.*;
+import com.cleanroommc.groovyscript.compat.mods.ModSupport;
+import com.cleanroommc.groovyscript.helper.recipe.AbstractRecipeBuilder;
+import com.cleanroommc.groovyscript.registry.StandardListRegistry;
+import org.jetbrains.annotations.Nullable;
+import sblectric.lightningcraft.api.recipes.LightningInfusionRecipe;
+import sblectric.lightningcraft.recipes.LightningInfusionRecipes;
+
+import java.util.Collection;
+
+@RegistryDescription
+public class Infusion extends StandardListRegistry<LightningInfusionRecipe> {
+
+    @Override
+    public Collection<LightningInfusionRecipe> getRecipes() {
+        return LightningInfusionRecipes.instance().getRecipeList();
+    }
+
+    @MethodDescription(example = @Example("item('minecraft:diamond')"))
+    public boolean removeByOutput(IIngredient output) {
+        return getRecipes().removeIf(r -> {
+            if (output.test(r.getOutput())) {
+                addBackup(r);
+                return true;
+            }
+            return false;
+        });
+    }
+
+    @RecipeBuilderDescription(example = {
+        @Example(".input(item('minecraft:clay'), item('minecraft:gold_ingot'), item('minecraft:gold_ingot'), item('minecraft:iron_ingot'), item('minecraft:iron_ingot')).output(item('minecraft:nether_star')).le(500)"),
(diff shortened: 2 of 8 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/mods/ModSupport.java
```diff
@@ -124,6 +125,7 @@ public class ModSupport {
     public static final GroovyContainer<Mekanism> MEKANISM = new InternalModContainer<>("mekanism", "Mekanism", Mekanism::new);
     public static final GroovyContainer<MysticalAgriculture> MYSTICAL_AGRICULTURE = new InternalModContainer<>("mysticalagriculture", "Mystical Agriculture", MysticalAgriculture::new);
     public static final GroovyContainer<LazyAE2> LAZYAE2 = new InternalModContainer<>("threng", "LazyAE2", LazyAE2::new, "lazyae2");
+    public static final GroovyContainer<Lightningcraft> LIGHTNINGCRAFT = new InternalModContainer<>("lightningcraft", "Lightningcraft", Lightningcraft::new);
```
Rongmario, line 128: `Lightningcraft` => `LightningCraft` (preferably change the class name to be the same case too)

src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        @Override
+        @RecipeBuilderRegistrationMethod
+        public @Nullable LightningInfusionRecipe register() {
+            if (!validate()) return null;
+            Object centerItem = input.get(0).getMatchingStacks()[0];
+            Object[] inputs = input.stream().skip(1).map(i -> i.getMatchingStacks()[0]).toArray();
+            LightningInfusionRecipe recipe = new LightningInfusionRecipe(output.get(0), le, centerItem, inputs);
+            if (nbtSensitive) recipe.setNBTSensitive();
```
Rongmario, line 100: Perhaps detect if inputs contain nbt tags instead of having a separate `nbtSensitive` call?

Wizzerinus: Or in addition to, because sometimes the dev might want the item that does not have NBT tags but have strict NBT matching.

src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+        }
+
+        @Override
+        @RecipeBuilderRegistrationMethod
+        public @Nullable LightningInfusionRecipe register() {
+            if (!validate()) return null;
+
+
+            Object centerItem = input.get(0).getMatchingStacks()[0];
```
WaitingIdly, line 101: i think this might deserve to be split off into a separate field

WaitingIdly (changes requested):
there is also lightning transforming recipes, according to MoreTweaker, although those might deserve a similar treatment to Pneu explosion (creation of native functionality + suggestion to use that instead of custom compat)

Wizzerinus:
I've heard the LC's own lightning transform is really buggy (see: DJ2 where you can only craft electricium in a specific order). Do not mind trying to implement lightning transform properly under in-world crafting.

src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
Wizzerinus: I originally had it as a separate field, but really it has the same treatment as other inputs, and the only difference is that it's a separate argument for the infusion table recipe record

brachy84: I would also prefer this to be separate

Wizzerinus: sure, will update.

Wizzerinus:
updated

src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
```diff
@@ -0,0 +1,133 @@
...
+                msg.add(centerItem.getAmount() > 1, "Expected stack size of 1 for {}, got {}", centerItem, centerItem.getAmount());
+            }
+        }
+
+        @Override
+        @RecipeBuilderRegistrationMethod
+        public @Nullable LightningInfusionRecipe register() {
+            if (!validate()) return null;
+
+            ItemStack centerItem = this.centerItem.getMatchingStacks()[0];
```
WaitingIdly, line 109: should loop for each of `getMatchingStacks()` instead of only taking the first.

src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
```diff
@@ -0,0 +1,133 @@
...
+            ItemStack[] inputs = input.stream().map(i -> i.getMatchingStacks()[0]).toArray(ItemStack[]::new);
```
WaitingIdly, line 110: while it is not currently on `main` and is still just in the Betweenlands PR, `IngredientHelper.cartesianProductItemStacks` would be very useful here to, once again, use all values of `getMatchingStacks()` instead of just the first.

src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
```diff
@@ -0,0 +1,133 @@
...
+            validateFluids(msg);
+            msg.add(centerItem == null, "Center item must not be null");
+            msg.add(centerItem != null && centerItem.getMatchingStacks().length == 0, "Center item must not have a matching item");
+            msg.add(le < 0, "LE cost must be positive");
+            for (IIngredient it : this.input) {
+                msg.add(it == null || it.getMatchingStacks().length == 0, "All inputs must have a matching item");
+            }
+            if (GroovyScriptConfig.compat.checkInputStackCounts && centerItem != null) {
+                msg.add(centerItem.getAmount() > 1, "Expected stack size of 1 for {}, got {}", centerItem, centerItem.getAmount());
+            }
```
WaitingIdly, line 101: use this instead
```suggestion
            msg.add(IngredientHelper.overMaxSize(centerItem, 1), "centerItem must have a stack size of 1");
```

src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
```diff
@@ -0,0 +1,133 @@
...
+    public static class RecipeBuilder extends AbstractRecipeBuilder<LightningInfusionRecipe> {
+
+        @Property()
+        private IIngredient centerItem = null;
+
+        @Property(comp = @Comp(gte = 0), defaultValue = "-1")
+        private int le = -1;
+
+        @Property(defaultValue = "determined automatically based on the input items")
+        private boolean nbtSensitive = false;
```
WaitingIdly, line 49: is there a reason why having this be directly configurable by the player is required? if it can be automatically determined, why not simply not expose it.

src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
```diff
@@ -0,0 +1,133 @@
...
+    })
+    public RecipeBuilder recipeBuilder() {
+        return new RecipeBuilder();
+    }
+
+    @Property(property = "input", comp = @Comp(gte = 0, lte = 4))
+    @Property(property = "output", comp = @Comp(eq = 1))
+    public static class RecipeBuilder extends AbstractRecipeBuilder<LightningInfusionRecipe> {
+
+        @Property()
```
WaitingIdly, line 42:
```suggestion
        @Property
```

WaitingIdly (changes requested):

src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
Wizzerinus: Oops, I forgot I made that helper method, wow.

src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
Wizzerinus: Because there are cases where it's automatically configured value is wrong. For example, if you want to allow only input item that has no NBT in it (say, it's some tool that stores durability in NBT). You need to enable nbtSensitive = true, and there's not really another way to do it

src/main/java/com/cleanroommc/groovyscript/compat/mods/lightningcraft/Infusion.java
Wizzerinus: Oop

Wizzerinus:
going to wait until betweenlands is merged for the cartesian product thing because I do not feel like reinventing the wheel and also because I feel like I'll have to rewrite to it anyway.

WaitingIdly:
the betweenlands PR has been merged, so now `IngredientHelper.cartesianProductItemStacks(List<IIngredient>)` can be used.

PR #284, opened 2025-01-03
add Blood Arsenal compat

WaitingIdly:
changes in this PR:
- add compat with Blood Arsenal `sanguineInfusion`. examples and i18n for that.


gradle.properties
```diff
@@ -27,6 +27,7 @@ debug_avaritia = false
 
 debug_better_with_mods = false
 debug_blood_magic = false
+debug_blood_arsenal = false
```
brachy84, line 30: a < m

brachy84 (changes requested):

gradle.properties
WaitingIdly: wow. i didnt know that. im hearing about this now for the first time.

brachy84: you are welcome

brachy84 (approved):

PR #275, opened 2024-12-03
Mod Object Mapper and Ingredient

WaitingIdly:
changes in this PR:
- add a mod Object Mapper. returns `ModContainer`.
- add `ModItemsIngredient` to get all ItemStacks associated with a given mod.
  - added cache functionality in there. not totally sure if its needed?

PR #289, opened 2025-03-14
Fallback property container for all installed mods

brachy84:
You now access all installed mods with `mods.[modid]`. There are basic getters for id, name, version and forge mod container.
`mods.[modid].allItems` returns a `IIngredient` with all items from that mod (which are exposed to the creative inventory).
Supersedes #275

### DIFF
--- a/examples/postInit/custom/vanilla.groovy
+++ b/examples/postInit/custom/vanilla.groovy
@@ -21,6 +21,10 @@ log.info(ore_iron >> (item_iron * 3)) // false
     println file.path
 }*/
 
+for (var stack in mods.theoneprobe.allItems) {
+    log.info stack
+}
+
 // Crafting recipes are typically created via recipe builder, but also have shorthand versions for some common uses.
 // Here are a series of examples, with the shorthand and corresponding recipe builder:
 
--- a/examples/postInit/custom/vanilla.groovy
+++ b/examples/postInit/custom/vanilla.groovy
@@ -21,7 +21,7 @@ log.info(ore_iron >> (item_iron * 3)) // false
     println file.path
 }*/
 
-for (var stack in mods.theoneprobe.allItems) {
+for (var stack in mods.minecraft.allItems[5..12]) {
     log.info stack
 }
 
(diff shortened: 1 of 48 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/mods/BasicGroovyPropertyContainer.java
```diff
@@ -0,0 +1,66 @@
...
+import net.minecraft.util.NonNullList;
+import net.minecraftforge.fml.common.Loader;
+import net.minecraftforge.fml.common.ModContainer;
+import net.minecraftforge.fml.common.registry.ForgeRegistries;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+public class BasicGroovyPropertyContainer {
```
WaitingIdly, line 17: `BasicGroovyPropertyContainer` isnt a great name. `Property` isnt even relevant here. `ModWrapper` perhaps? would suggest `ModContainer` except for the obvious.

examples/postInit/custom/vanilla.groovy
```diff
@@ -21,6 +21,10 @@ log.info(ore_iron >> (item_iron * 3)) // false
     println file.path
 }*/
 
+for (var stack in mods.theoneprobe.allItems) {
+    log.info stack
+}
```
WaitingIdly, line 26: while `theoneprobe` is included in the development environment, this file is for vanilla exclusive stuff - what about this? using `ItemStack#getTranslationKey` to make what is being printed more readable.
```suggestion
for (var stack in mods.minecraft.allItems.getItemStacks()[5..12]) {
    log.info stack.getTranslationKey()
}
```
side note: a bit surprised that `ItemsIngredient` doesnt implement `List` (it does implement `Iterable`)

WaitingIdly (changes requested):
a likely unexpected side effect of this is `mods.minecraft` now exists. i think this is the correct behavior, it just becomes a bit weird to speak of.
due to this, though, it makes sense to bind the currently existing vanilla compat to `mods.minecraft` - and bind the custom original groovyscript to `mods.groovyscript`.
additionally, it seems like it would make sense to add `compat` as an alias to `mods` (ie `compat.minecraft`).

src/main/java/com/cleanroommc/groovyscript/compat/mods/MinecraftModContainer.java
```diff
@@ -0,0 +1,60 @@
...
+    private final Collection<String> aliases;
+
+    MinecraftModContainer() {
+        this.modProperty = Suppliers.memoize(() -> {
+            VanillaModule t = VanillaModule.INSTANCE;
+            t.addPropertyFieldsOf(t, false);
+            return t;
+        });
+        Set<String> aliasSet = new ObjectOpenHashSet<>();
+        aliasSet.add("mc");
```
WaitingIdly, line 27: would `vanilla` be a good alias? would allow `mods.vanilla`, which i find rather amusing.
i believe using `aliasSet.addAll(Alias.generateOf("Vanilla", containerName).and("MC", "mc"));` should work.

src/main/java/com/cleanroommc/groovyscript/compat/mods/MinecraftModContainer.java
```diff
@@ -0,0 +1,60 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+
+public final class MinecraftModContainer extends GroovyContainer<VanillaModule> {
+
+    private final String modId = "minecraft";
+    private final String containerName = "Minecraft";
```
WaitingIdly, line 16: is there a reason not to have these be `static`?

WaitingIdly (approved):
looks good!

PR #295, opened 2025-04-02
Add reflection helpers as metaclass expansion

brachy84:
- adds `makePublic` and `makeMutable` methods to metaclass class


examples/postInit/reflection.groovy
WaitingIdly, line 1: could this be moved to the `custom` folder?

WaitingIdly (approved):

examples/postInit/reflection.groovy
brachy84: Huh, i thought i did that

PR #306, opened 2025-05-05
Fix Java 24 incompatibility

brachy84:
The reason for the crash is that Groovy uses its own version of ASM to read classes. In that version it can only read class files up to version 67 (java 23). Afaik only the classes from java itself have class version 68. 
This pr obtains the class bytes not by manually searching pushing an input stream through a class reader, but by using forge methods. Which for some reason doesnt return anything for java classes. I have no idea where the java classes come from because they obviously work.

This likely will become a problem again once mods start to compile for java 24 and higher. At that point we will have to pull to a few classes from groovys asm decompiler and adjust them.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/GroovyScript.java
+++ b/src/main/java/com/cleanroommc/groovyscript/GroovyScript.java
@@ -101,7 +101,7 @@ public class GroovyScript {
 
     @Mod.EventHandler
     public void onConstruction(FMLConstructionEvent event) {
-        JavaVersionCheck.validateJavaVersion(event.getSide());
+        //JavaVersionCheck.validateJavaVersion(event.getSide());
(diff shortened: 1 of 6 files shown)


Rongmario:
- Startup needed classes (essential Java classes, rt.jar and other jre/lib jars) gets loaded in the `null` class loader, basically the `BootstrapClassLoader`.
- Other Java/JDK classes gets loaded in the `PlatformClassLoader`.
- `AppClassLoader` or `SystemClassLoader` is what comes next, can be overriden by a startup argument, which is what `Cleanroom` does for `Foundation`'s `ActualClassLoader`'s parent.

kappa-maintainer:
I no longer set launch class loader in argument now, but `Foundation` still supports doing so

PR #313, opened 2025-05-21
feat: Add author field to RunConfig

Xikaro:
Changes:
- Add packAuthor field to RunConfig

### DIFF
--- a/examples/runConfig.json
+++ b/examples/runConfig.json
@@ -1,28 +1,22 @@
 {
-  "packName": "GroovyScript Dev",
-  "packId": "groovyscriptdev",
-  "author": "Placeholder author",
+  "packName": "PlaceHolder name",
+  "packId": "placeholdername",
+  "author": "Placeholder, author",
   "version": "1.0.0",
-  "debug": true,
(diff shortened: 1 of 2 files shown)


brachy84 (changes requested):
Please revert the other changes in runConfig

examples/runConfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "packName": "GroovyScript Dev",
   "packId": "groovyscriptdev",
+  "author": "GroovyScript, Dev, Author",
```
WaitingIdly, line 4: should be an array - `["GroovyScript", "Dev", "Author"]`

Xikaro: No, I would like to see it as a regular string, and then inside it is parsed as needed for modMetadata, like the curseforge manifest, pakku
Screenshot_633

WaitingIdly: thats CurseForge's format - for mods, in the `mcmod.info` the relevant element is `authorList`, an array. this is creating a mod, and so it should be an array.

PR #325, opened 2025-08-28
Note Chisel Carving bug for registry replacement

WaitingIdly:
changes in this PR:
- add a warning about registry replacement breaking chisel and how its probably not our fault please blame some other mod (so far, just Inspirations is at fault).
- add this warning to the wiki.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/compat/mods/chisel/Carving.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/mods/chisel/Carving.java
@@ -98,6 +98,8 @@ public class Carving extends VirtualizedRegistry<Pair<String, ItemStack>> {
                     .add("instead, edit the oredict via `oredict.remove('{}', {})`", groupName, GroovyScriptCodeConverter.asGroovyCode(item, false, false))
                     .error()
                     .post();
+        } catch (NullPointerException e) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/mods/chisel/Carving.java
```diff
@@ -95,6 +98,8 @@ public void removeVariation(String groupName, ItemStack item) {
                     .add("instead, edit the oredict via `oredict.remove('{}', {})`", groupName, GroovyScriptCodeConverter.asGroovyCode(item, false, false))
                     .error()
                     .post();
+        } catch (NullPointerException e) {
+            GroovyLog.get().exception("An exception occurred with chisel carving - possibly due to some other mod doing registry replacement.", e);
```
brachy84, line 102: Is there something the user can do to fix the crash? If no then it should just be an error. (And mention what the user is else supposed to do). Otherwise still display the fix message.

brachy84 (changes requested):

src/main/java/com/cleanroommc/groovyscript/compat/mods/chisel/Carving.java
brachy84: Then the user is supposed to remove what they are currently doing? The message should say that,

PR #338, opened 2025-10-04
Backwards compat for classes property

brachy84:
Previously the game would just crash since 1.2.3 when classes were defined in the runconfig. Now only an error will be logged and the paths will be added to the loaders automatically.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/sandbox/RunConfig.java
+++ b/src/main/java/com/cleanroommc/groovyscript/sandbox/RunConfig.java
@@ -141,11 +141,35 @@ public class RunConfig {
 
         String regex = File.separatorChar == '\\' ? "/" : "\\\\";
         String replacement = getSeparator();
-        if (json.has("classes")) {
-            throw new IllegalStateException("GroovyScript classes definition in runConfig is deprecated! Classes are now treated as normal scripts.");
+
+        JsonElement el = json.get("loaders");
+        JsonObject jsonLoaders;
+        if (el == null || !el.isJsonObject()) {
+            GroovyLog.msg("No loaders are defined (or is not a json object). This means no scripts will be executed.")
+                    .add("Please see https://cleanroommc.com/groovy-script/getting_started/run_config#loaders for help.")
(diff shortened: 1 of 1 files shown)


src/main/java/com/cleanroommc/groovyscript/sandbox/RunConfig.java
```diff
@@ -141,11 +141,35 @@ public void reload(JsonObject json, boolean init) {
 
         String regex = File.separatorChar == '\\' ? "/" : "\\\\";
         String replacement = getSeparator();
+
+        JsonElement el = json.get("loaders");
+        JsonObject jsonLoaders;
+        if (el == null || !el.isJsonObject()) {
+            GroovyLog.msg("No loaders are defined (or is not a json object). This means no scripts will be executed.")
```
WaitingIdly, line 148: due to how GroovyLog formats this, this appears in the log as this:
```log
[Client thread/ERROR] [GroovyLog]: GroovyScript classes definition in runConfig is deprecated! Classes are now treated as normal scripts.: 
[Client thread/ERROR] [GroovyLog]:  - GroovyScript will try to add the defined paths to the loaders automatically. This may result in unexpected behaviour.
[Client thread/ERROR] [GroovyLog]:  - Visit https://cleanroommc.com/groovy-script/getting_started/run_config#classes to find out how to migrate.
[Client thread/ERROR] [GroovyLog]:  - Ask on the discord if you need more help.
```
it seems like printing multiple separate lines would be better, but if thats not desired the period at the end of the first line should at least be removed.
```java
GroovyLog.get().errorMC("GroovyScript classes definition in runConfig is deprecated! Classes are now treated as normal scripts.");
GroovyLog.get().errorMC("GroovyScript will try to add the defined paths to the loaders automatically. This may result in unexpected behaviour.");
GroovyLog.get().errorMC("Visit https://cleanroommc.com/groovy-script/getting_started/run_config#classes to find out how to migrate.");
GroovyLog.get().errorMC("Ask on the discord if you need more help.");
```

PR #341, opened 2025-10-07
feat: ProjectE EMC Mapper

Amai-ICE:
Sorry, I use Google Translate
Crafting recipes created with GroovyScript do not perform EMC calculations, so this feature corrects that.
The calculation is applicable to recipes that implement ShapedCraftingRecipe and ShapelessCraftingRecipe.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/compat/mods/projecte/EMCMapper.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/mods/projecte/EMCMapper.java
@@ -1,0 +1,24 @@
+package com.cleanroommc.groovyscript.compat.mods.projecte;
+
+import com.cleanroommc.groovyscript.compat.vanilla.ShapedCraftingRecipe;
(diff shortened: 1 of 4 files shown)


brachy84:
surely project e has an api for it and we dont have to use mixin

Rongmario:
It is how ProjectE does with CraftTweaker recipes... Very unconventional, but other than reflection, mixin is the only way.

See: [CraftTweakerRecipeMapper](https://github.com/sinkillerj/ProjectE/blob/mc1.12.x/src/main/java/moze_intel/projecte/emc/mappers/CraftTweakerRecipeMapper.java) and [CraftingMapper](https://github.com/sinkillerj/ProjectE/blob/4773d4197cfc60478749c7dea49748d7c2f28be8/src/main/java/moze_intel/projecte/emc/mappers/CraftingMapper.java#L34)

brachy84:
Peak non existing api design

PR #351, opened 2025-10-21
Reload JEIless

WaitingIdly:
changes in this PR:
- add the ability to run `/gs reload --skip-jei` to not reload JEI.
- add a keybind "Quick Reload" (unbound by default) to only reload scripts, not JEI.
- attempt to run `JEIStarter#load` if it exists, otherwise fall back to `JEIStarter#start`. allows faster reloading with HEI.
- improve the jei reload message appearance and functionality, including suggesting HEI if not installed and clicking to reload everything if only recipes were reloaded.
- add `StyleConstant.getTipStyle()` for emphasized text that isnt based on status and often has interaction.

this will help script testing time, as reloading JEI is >90% of the time it takes to reload in many large packs.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/helper/StyleConstant.java
+++ b/src/main/java/com/cleanroommc/groovyscript/helper/StyleConstant.java
@@ -67,7 +67,7 @@ public class StyleConstant {
      * Used for titles to make them emphasised and keep them distinct from the surrounding text
      */
     public static Style getTitleStyle() {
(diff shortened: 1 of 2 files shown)


IntegerLimit:
I think if this PR could also include a flag to reload JEI recipes only (and not ingredients), that would be nice, as that provides basic JEI reloadability at a much quicker speed.

HEI has in-built compat for this; perhaps this feature can be HEI only? Or, GrS can implement their own version so that it can be used in both JEI and HEI.

brachy84 (approved):

PR #381, opened 2026-04-11
fix: Factory Tech issues

Wizzerinus:
This PR fixes the following issues with Factory Tech compatibility:

- Removes Circuit Scribe integration, because it does not actually work. The tile entity determines if Circuit Scribe can work by comparing its input item with `circuit_intermediate:0` and its output item with `circuit_intermediate:X`, which means custom recipes can't be added. (see https://gitlab.com/dalapo/FactoryTech/-/blob/master/src/main/java/dalapo/factech/tileentity/specialized/TileEntityCircuitScribe.java)
- Removes `getMaxItemInput` from the machines that properly handle item input size. These machines are `Centrifuge, Magnet Centrifuge, Grindstone, High Tech Furnace, Magnetizer, Metal Cutter, Ore Drill, Reclaimer, Saw`. The other machines fail in different ways when item input size != 1 (either requiring X but only consuming 1 item, consuming X but running with < X, or both).


src/main/java/com/cleanroommc/groovyscript/compat/mods/factorytech/FactoryTech.java
WaitingIdly, line 1: maybe make a comment about the circuit scribe?

WaitingIdly (approved):

src/main/java/com/cleanroommc/groovyscript/compat/mods/factorytech/FactoryTech.java
Wizzerinus: done

brachy84 (approved):

PR #382, opened 2026-04-11
Refactor JEI description, add support for fluids

Wizzerinus:
Allows using `fluid(...)` in `mods.jei.description`, which also fixes a startup crash when the code does that

This should support adding more types fairly easily

lmk if this is a bad approach, I was kinda very focused on keeping it a VirtualizedRegistry, and its probably not the cleanest way

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
@@ -2,7 +2,6 @@ package com.cleanroommc.groovyscript.compat.mods.jei;
 
 import java.util.Arrays;
 import java.util.List;
-import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -16,69 +15,72 @@ import com.cleanroommc.groovyscript.api.documentation.annotations.Example;
 import com.cleanroommc.groovyscript.api.documentation.annotations.MethodDescription;
 import com.cleanroommc.groovyscript.api.documentation.annotations.RegistryDescription;
 import com.cleanroommc.groovyscript.core.mixin.jei.IngredientInfoRecipeAccessor;
-import com.cleanroommc.groovyscript.helper.ingredient.ItemsIngredient;
 import com.cleanroommc.groovyscript.registry.VirtualizedRegistry;
 
 import mezz.jei.api.IModRegistry;
 import mezz.jei.api.IRecipeRegistry;
-import mezz.jei.api.ingredients.VanillaTypes;
+import mezz.jei.api.ingredients.IIngredientHelper;
+import mezz.jei.api.ingredients.IIngredientRegistry;
 import mezz.jei.api.recipe.IIngredientType;
 import mezz.jei.api.recipe.VanillaRecipeCategoryUid;
 import mezz.jei.plugins.jei.info.IngredientInfoRecipeCategory;
-import net.minecraft.item.ItemStack;
 import net.minecraftforge.fluids.FluidStack;
 
 @RegistryDescription(category = RegistryDescription.Category.ENTRIES)
-public class Description extends VirtualizedRegistry<Description.DescriptionEntry<?>> {
+public class Description extends VirtualizedRegistry<Description.DescriptionEntry> {
 
-    static class DescriptionEntry<T> {
-        T representative;
-        List<T> ingredients;
-        IIngredientType<T> ingredientType;
+    public static class DescriptionEntry {
+        List<Object> ingredients;
         String[] descriptionKeys;
 
-        DescriptionEntry(@Nonnull List<T> ingredients, @Nonnull IIngredientType<T> ingredientType, @Nullable List<String> descriptionKeys) {
+        @SuppressWarnings("unchecked")
+        DescriptionEntry(@Nonnull List<?> ingredients, @Nullable List<String> descriptionKeys) {
             if (ingredients.isEmpty()) {
                 throw new IllegalArgumentException("JEI Description Entries must not have an empty list of ingredients, got " + ingredients);
             }
-            Objects.requireNonNull(ingredientType);
 
-            this.ingredients = ingredients;
-            this.ingredientType = ingredientType;
-            this.representative = this.ingredients.get(0);
-            Objects.requireNonNull(representative);
+            // This cast is not exactly correct, but Java/JVM allows it due to type erasure
+            this.ingredients = (List<Object>) ingredients;
 
             this.descriptionKeys = descriptionKeys == null ? new String[0] : descriptionKeys.toArray(new String[0]);
         }
+
+        @Nullable
+        IIngredientType<Object> validateIngredientType(IModRegistry modRegistry) {
+            Object first = ingredients.get(0);
+            IIngredientType<Object> ingredientType = modRegistry.getIngredientRegistry().getIngredientType(first);
+            if (ingredientType == null) {
+                return null;
+            }
+            for (Object o : ingredients) {
+                if (!ingredientType.equals(modRegistry.getIngredientRegistry().getIngredientType(o))) {
+                    return null;
+                }
+            }
+            return ingredientType;
+        }
+    }
+
+    private <T> void addIngredientInfo(IModRegistry modRegistry, List<T> ingredients, IIngredientType<T> ingredientType, String[] description) {
+        // Force Java to pick the correct overload we use
+        // This is needed because the method has overloads in a way T = Object cannot be substituted
+        modRegistry.addIngredientInfo(ingredients, ingredientType, description);
     }
 
     /**
      * Called by {@link JeiPlugin#register}
      */
-    @SuppressWarnings("unchecked")
     @GroovyBlacklist
     public void applyAdditions(IModRegistry modRegistry) {
-        for (DescriptionEntry<?> entry : this.getScriptedRecipes()) {
-            if (entry.representative instanceof ItemStack) {
-                modRegistry.addIngredientInfo((List<ItemStack>) entry.ingredients, (IIngredientType<ItemStack>) entry.ingredientType, entry.descriptionKeys);
-            } else if (entry.representative instanceof FluidStack) {
-                modRegistry.addIngredientInfo((List<FluidStack>) entry.ingredients, (IIngredientType<FluidStack>) entry.ingredientType, entry.descriptionKeys);
+        for (DescriptionEntry entry : this.getScriptedRecipes()) {
+            IIngredientType<Object> ingredientType = entry.validateIngredientType(modRegistry);
+            if (ingredientType != null) {
+                addIngredientInfo(modRegistry, entry.ingredients, ingredientType, entry.descriptionKeys);
             } else {
-                GroovyLog.msg("Got a Description Entry of an invalid type, this is probably a bug in Groovyscript").error().post();
+                GroovyLog.msg("Unable to obtain an ingredient type for items {}", entry.ingredients.toArray())
+                    .error()
+                    .post();
             }
-        }
-    }
-
-    private boolean compareObjects(Object inEntry, Object inRegistry) {
-        if (inEntry instanceof ItemStack ieStack) {
-            return inRegistry instanceof ItemStack irStack && new ItemsIngredient(ieStack).test(irStack);
-        } else if (inEntry instanceof FluidStack ieStack) {
-            return inRegistry instanceof FluidStack irStack && ieStack.getFluid().equals(irStack.getFluid());
-        } else {
-            GroovyLog.msg("Got a Description Entry of type {} which is not supported, this is probably a bug in Groovyscript", inEntry.getClass().getName())
-                .error()
-                .post();
-            return false;
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -1,60 +1,107 @@
...
     @GroovyBlacklist
     public void applyAdditions(IModRegistry modRegistry) {
-        for (Pair<List<IIngredient>, List<String>> entry : this.getScriptedRecipes()) {
-            modRegistry.addIngredientInfo(
-                    entry.getLeft().stream().flatMap(x -> Stream.of(x.getMatchingStacks())).collect(Collectors.toList()),
-                    // Currently, it is only possible to add VanillaTypes.ITEM. It may be desirable to add the ability to do other types.
-                    VanillaTypes.ITEM,
-                    entry.getRight().toArray(new String[0]));
+        for (DescriptionEntry<?> entry : this.getScriptedRecipes()) {
+            System.out.println("adding entry: " + entry);
```
WaitingIdly: println

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -1,60 +1,107 @@
...
             recipeRegistry.getRecipeWrappers(category).forEach(wrapper -> {
                 IngredientInfoRecipeAccessor<?> accessor = (IngredientInfoRecipeAccessor<?>) wrapper;
 
-                // Currently, it is only possible to remove VanillaTypes.ITEM. It may be desirable to add the ability to do other types.
-                if (!VanillaTypes.ITEM.equals(accessor.getIngredientType())) return;
-
-                for (Pair<List<IIngredient>, List<String>> entry : this.getBackupRecipes()) {
-                    if (entry.getKey()
+                for (DescriptionEntry<?> entry : this.getBackupRecipes()) {
+                    System.out.println("removing entry: " + entry);
```
WaitingIdly: println

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -69,32 +116,58 @@ public void onReload() {
     }
 
     @MethodDescription(type = MethodDescription.Type.ADDITION)
-    public void add(List<IIngredient> target, List<String> description) {
-        addScripted(Pair.of(target, description));
+    public void add(IIngredient[] target, String... description) {
```
WaitingIdly: maybe do `Object[] target` + get the `IIngredientRegistry#getIngredientType` later on + check that everything has the same ingredient type. wouldnt use `IIngredient` or `FluidStack` anymore

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -1,60 +1,107 @@
...
+            } else if (entry.representative instanceof FluidStack) {
+                modRegistry.addIngredientInfo((List<FluidStack>) entry.ingredients, (IIngredientType<FluidStack>) entry.ingredientType, entry.descriptionKeys);
+            } else {
+                GroovyLog.msg("Got a Description Entry of an invalid type, this is probably a bug in Groovyscript").error().post();
+            }
+        }
+    }
+
+    private boolean compareObjects(Object inEntry, Object inRegistry) {
+        if (inEntry instanceof ItemStack ieStack) {
```
WaitingIdly: since this happens only in `applyRemovals`, we can compare using `IIngredientHelper#getUniqueId` - perhaps `IngredientUtil#equal`? although i think that might be intended for internal logic.

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -1,60 +1,107 @@
...
     public void applyAdditions(IModRegistry modRegistry) {
-        for (Pair<List<IIngredient>, List<String>> entry : this.getScriptedRecipes()) {
-            modRegistry.addIngredientInfo(
-                    entry.getLeft().stream().flatMap(x -> Stream.of(x.getMatchingStacks())).collect(Collectors.toList()),
-                    // Currently, it is only possible to add VanillaTypes.ITEM. It may be desirable to add the ability to do other types.
-                    VanillaTypes.ITEM,
-                    entry.getRight().toArray(new String[0]));
+        for (DescriptionEntry<?> entry : this.getScriptedRecipes()) {
+            System.out.println("adding entry: " + entry);
+            if (entry.representative instanceof ItemStack) {
```
WaitingIdly: would suggest ensuring everything is the same `IIngredientRegistry#getIngredientType` + registering via that (casting as needed. shouldnt need an `instanceof` then

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -1,60 +1,107 @@
...
-import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import net.minecraftforge.fluids.FluidStack;
 
 @RegistryDescription(category = RegistryDescription.Category.ENTRIES)
-public class Description extends VirtualizedRegistry<Pair<List<IIngredient>, List<String>>> {
+public class Description extends VirtualizedRegistry<Description.DescriptionEntry<?>> {
+
+    static class DescriptionEntry<T> {
```
WaitingIdly: think this needs `public`

WaitingIdly (changes requested):

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
Wizzerinus: The problem is sometimes you need to remove description from an item with a specific NBT, which checking only the ID doesn't. I wanted to keep the old logic in regards to this. Though I don't really like how this turned out, yeah

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
Wizzerinus: As of right now it's only used within the class, any reason why?

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
Wizzerinus: 1 am coding :heart:

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
Wizzerinus: Hmm, we could do that I guess, that would be much better in terms of future flexibility too

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
Wizzerinus: This is not possible due to a Java quirk, that I have already outlined in the Discord channel. The JEI API exposes an overloaded function like this:

```
<T> void addIngredientInfo(T ingredient, IIngredientType<T> ingredientType, String... descriptionKeys);
<T> void addIngredientInfo(List<T> ingredients, IIngredientType<T> ingredientType, String... descriptionKeys);
```

If `T := Object` is substituted (or `T` being an existential type, which eventually reduces to Object), then these two methods are ambiguous, so `T` must be a concrete type, hence this is necessary.

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
Wizzerinus: See my comment on `applyAdditions`

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
WaitingIdly: part of the signature, `Description extends VirtualizedRegistry<Description.DescriptionEntry<?>>`

Wizzerinus: is this for consistency reasons? it works fine as is, but if its for consistency idm fixing

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
brachy84: Why is this changed to array? Lists are much nicer in groovy.

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -69,32 +119,80 @@ public void onReload() {
...
+    public void add(IIngredient[] target, String... description) {
+        addScripted(new DescriptionEntry(
+            Stream.of(target).flatMap(x -> Stream.of(x.getMatchingStacks())).collect(Collectors.toList()),
+            Arrays.asList(description)));
     }
 
     @MethodDescription(type = MethodDescription.Type.ADDITION)
-    public void add(List<IIngredient> target, String... description) {
-        addScripted(Pair.of(target, Arrays.asList(description)));
+    public void add(IIngredient[] target, List<String> description) {
```
brachy84: This too

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -69,32 +119,80 @@ public void onReload() {
...
 
-    @MethodDescription(type = MethodDescription.Type.ADDITION, example = @Example("item('minecraft:gold_ingot'), 'groovyscript.recipe.fluid_recipe'"))
+    @MethodDescription(type = MethodDescription.Type.ADDITION, example = @Example("item('minecraft:gold_ingot') | item('minecraft:iron_block'), 'groovyscript.recipe.fluid_recipe'"))
     public void add(IIngredient target, String... description) {
-        addScripted(Pair.of(Collections.singletonList(target), Arrays.asList(description)));
+        add(new IIngredient[]{target}, description);
+    }
+
+    @MethodDescription(type = MethodDescription.Type.ADDITION)
+    public void add(Object[] target, String... description) {
```
brachy84: List > array

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -69,32 +119,80 @@ public void onReload() {
...
+        addScripted(new DescriptionEntry(
+            Arrays.asList(target),
+            Arrays.asList(description)));
+    }
+
+    @MethodDescription(type = MethodDescription.Type.ADDITION)
+    public void add(Object[] target, List<String> description) {
```
brachy84: List > array

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -69,32 +119,80 @@ public void onReload() {
...
+
+    @MethodDescription(type = MethodDescription.Type.ADDITION)
+    public void add(Object target, String... description) {
+        add(new Object[]{target}, description);
+    }
+
+    // This overload set is the same as Object one. It exists because FluidStack implements IIngredient, but does not have any matching stacks,
+    // which causes a JEI startup crash when trying to add that description
+    @MethodDescription(type = MethodDescription.Type.ADDITION)
+    public void add(FluidStack[] target, String... description) {
```
brachy84: List > array

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -69,32 +119,80 @@ public void onReload() {
...
+        addScripted(new DescriptionEntry(
+            Arrays.asList(target),
+            Arrays.asList(description)));
+    }
+
+    @MethodDescription(type = MethodDescription.Type.ADDITION)
+    public void add(FluidStack[] target, List<String> description) {
```
brachy84: List > array

brachy84 (changes requested):

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
Wizzerinus: not possible due to type erasure

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
Wizzerinus: because then we cannot overload based on the input type, as all lists cast down to `List<?>` on the JVM, and we need to apply different code depending on FluidStack, an IIngredient that is not FluidStack, or another type (the latter is currently not supported in JEI so idk an example for that). We *could* disambiguate manually though. Idk, these group methods are very uncommon I believe (since for IIngredient you can just use an OrIngredient), so it's hard to tell whether the manual overloading will be worth in the long run.

Wizzerinus:
refactored to use lists

src/main/java/com/cleanroommc/groovyscript/compat/mods/jei/Description.java
```diff
@@ -68,33 +119,51 @@ public void onReload() {
...
+            GroovyLog.msg("The ingredient list cannot be empty")
+                .error()
+                .post();
+            return null;
+        }
+
+        return target.stream()
+            .<Object> flatMap(x ->
+                x instanceof IIngredient && !(x instanceof FluidStack)
+                ? Stream.of(((IIngredient)x).getMatchingStacks()) : Stream.of(x))
```
WaitingIdly: in what situation would this be called? if its `item('whatever')` then its already an itemstack - no need to convert - and if its `ore('whatever')` it should just be rejected. this code means a `gas('whatever')` would be converted into an `ItemStack` stream (it is an `IIngredient`).

Wizzerinus: Why do we need to reject ore(whatever)?

I didn't think of the gas implications though, that one is fair. Maybe we add a new method to IIngredient that determines whether this ingredient handles items or not

WaitingIdly: > Why do we need to reject ore(whatever)?

because jei uses IngredientTypes, only things that are of an IngredientType is relevant. `ore('whatever')` doesnt create one of these

PR #389, opened 2026-08-18
Ice & Fire RLCraft support

CaliforniaDemise:
Rlcraft edition moves recipe lists to a different class which makes it incompatible with GroovyScript. This PR fixes it adding list parameter to constructor and getting the right field via checking for the version in mods container.

### DIFF
--- a/src/main/java/com/cleanroommc/groovyscript/compat/mods/iceandfire/DragonForge.java
+++ b/src/main/java/com/cleanroommc/groovyscript/compat/mods/iceandfire/DragonForge.java
@@ -12,9 +12,11 @@ import java.util.Collection;
 
 public abstract class DragonForge extends StandardListRegistry<DragonForgeRecipe> {
 
+    private final String type;
     private final Collection<DragonForgeRecipe> recipes;
 
-    public DragonForge(Collection<DragonForgeRecipe> recipes) {
+    protected DragonForge(String type, Collection<DragonForgeRecipe> recipes) {
+        this.type = type;
         this.recipes = recipes;
     }
 
@@ -41,7 +43,7 @@ public abstract class DragonForge extends StandardListRegistry<DragonForgeRecipe
 
         @Override
         public String getErrorMsg() {
-            return "Error adding Ice And Fire Fire Forge recipe";
+            return "Error adding Ice And Fire " + DragonForge.this.type + " Forge recipe";
         }
 
         @Override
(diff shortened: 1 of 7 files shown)


src/main/java/com/cleanroommc/groovyscript/compat/mods/iceandfire/IceAndFire.java
```diff
@@ -1,22 +1,48 @@
...
     }
 
+    private static Version version() {
+        var entry = Loader.instance().getIndexedModList().get("iceandfire");
+        if (entry == null) return Version.ORIGINAL;
+        // Name should be "Ice And Fire: RotN Edition"
+        if (entry.getName().contains("RotN")) return Version.ROTN;
+        // Ice And Fire 2.x most likely means RLCraft edition
+        else if (entry.getVersion().startsWith("2")) return Version.RLCRAFT;
+        else return Version.ORIGINAL;
```
WaitingIdly: dont need `else` if you just returned

src/main/java/com/cleanroommc/groovyscript/compat/mods/iceandfire/IceForge.java
```diff
@@ -1,72 +1,38 @@
...
     }
 
     @RecipeBuilderDescription(example = {
             @Example(".input(item('minecraft:gold_ingot'), item('minecraft:gold_ingot')).output(item('minecraft:clay'))"),
             @Example(".input(item('minecraft:diamond'), item('minecraft:gold_ingot')).output(item('minecraft:clay'))")
     })
+    @Override
     public RecipeBuilder recipeBuilder() {
-        return new RecipeBuilder();
+        return super.recipeBuilder();
```
WaitingIdly: you can move these various annotations to `@RegistryDescription` - see `@MethodOverride` - and remove the stub methods

src/main/java/com/cleanroommc/groovyscript/compat/mods/iceandfire/DragonForge.java
```diff
@@ -0,0 +1,67 @@
...
+        return getRecipes().removeIf(r -> output.test(r.getOutput()) && doAddBackup(r));
+    }
+
+    @Property(property = "input", comp = @Comp(eq = 2))
+    @Property(property = "output", comp = @Comp(eq = 1))
+    public final class RecipeBuilder extends AbstractRecipeBuilder<DragonForgeRecipe> {
+
+        @Override
+        public String getErrorMsg() {
+            return "Error adding Ice And Fire Fire Forge recipe";
```
WaitingIdly: this claims to always be "Fire Forge", which is not correct

